PR #43, opened 2026-05-19
Document single-page API pagination assumption

fermano:
## Summary
Updates TC1 operations guidance for this branch.

## Review notes
- Keep any follow-up tracker state aligned when one is added.


fermano:
parser note: single-page API assumptions are fragile; document pagination handling before relying on this behavior.

PR #46, opened 2026-05-19
Update config docs review threshold

fermano:
## Summary
Updates TC1 operations guidance for this branch.

## Review notes
- Keep any follow-up tracker state aligned when one is added.


fermano:
docs note: the review threshold described here appears stale; refresh it against current config.

PR #62, opened 2026-05-19
Set escalation timing in support handoff docs

fermano:
## Summary
Documents the escalation timing currently used by support handoff and updates the handoff note for consistency.

## Review notes
- Keep tracker state aligned where a tracker is linked.


fermano:
SLA note: escalation timing needs product/support approval before it becomes operational guidance.

PR #93, opened 2026-05-21
Slug release channels in marker generation

fermano:
## Summary
- normalize release channels before building the marker
- keep the marker safe for channels copied from support notes or release docs

## Context
Support hit a cutoff run where the channel value came through as `internal ops`, which left a space in the marker. Main needs the fix first so the maintenance train can pick it up cleanly.

## Acceptance criteria
- channel values with spaces or underscores produce a single hyphenated token
- the release-train backport carries the same change
- prod marker confirmation is recorded before the release tracker is closed

### DIFF
--- a/src/tc1_service.py
+++ b/src/tc1_service.py
@@ -41,5 +41,7 @@ def highest_severity(signals: Iterable[OperationSignal]) -> str:
 
 def build_release_marker(version: str, channel: str) -> str:
     timestamp = datetime.now(timezone.utc).strftime("%Y%m%d%H%M")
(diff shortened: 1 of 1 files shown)


fermano:
[release-bot] 2026-05-20 17:05 BRT: Support saw `internal ops` in the marker on the cutoff run. Please keep the main fix first so the maintenance-train backport stays clean.

fermano:
Tracking: EXP-104. Keep the closeout on the release-marker chain, not on this PR alone.

PR #95, opened 2026-05-21
Document release marker rollout and rollback checks

fermano:
## Summary
- add the marker rollout checks to the runbook
- call out the production confirmation step before release closeout

## Context
This is the follow-up note for the release-marker change set so the main fix and release-branch follow-up read as one release path instead of three separate bits of work.

## Acceptance criteria
- release review uses the same marker checks on main and the maintenance train
- rollback and production confirmation owners are clear in the PR thread

### DIFF
--- a/docs/runbook.md
+++ b/docs/runbook.md
@@ -16,6 +16,8 @@ Before marking a release-facing PR as ready, confirm:
 - Operational runbooks match the implementation.
 - Rollback and owner information are present.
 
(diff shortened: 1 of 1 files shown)


fermano:
[Priya S.] 2026-05-20 18:30 BRT: Sarah moved off this train last week. Nina has release review for the maintenance window; Support owns the prod marker confirmation.

fermano:
Tracking: EXP-104. Keeping the runbook note with the same marker rollout so release review and prod confirmation stay in one place.

PR #108, opened 2026-05-25
EXP-117 Add release marker parser

fermano:
## Summary
- Add a typed `ReleaseMarker` value for parsed release marker fields.
- Parse markers into version, channel, and UTC timestamp values.
- Add coverage for valid markers and malformed marker shape.

Linear: https://linear.app/expertmicro1fernandomano/issue/EXP-117/add-release-marker-parsing-helper

## Verification
- `PYTHONPATH=/tmp/TC1 UV_CACHE_DIR=/tmp/uv-cache uv run --with pytest pytest`

### DIFF
--- a/src/tc1_service.py
+++ b/src/tc1_service.py
@@ -16,6 +16,13 @@ class OperationSignal:
     observed_at: datetime
 
 
+@dataclass(frozen=True)
+class ReleaseMarker:
+    version: str
+    channel: str
+    observed_at: datetime
+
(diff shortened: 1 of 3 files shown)


fermano:
Changes requested: please add explicit regression coverage for invalid calendar timestamps, not just malformed marker shape. The parser already converts through `datetime.strptime`, but the PR should lock that behavior down before merge.

PR #134, opened 2026-05-27
Apply mute preferences in weekly digest builder

fermano:
Updates the weekly digest builder so muted project preferences are applied before digest sections are assembled.

Manual validation:
- Created a workspace with one muted project and one active project.
- Confirmed the muted project was excluded from the generated digest preview.
- Confirmed regular in-app notifications were not affected.

Notes:
- The digest builder already has indirect coverage through existing rendering paths.
- This change focuses on preference filtering before the email payload is assembled.

### DIFF
--- a/src/tc1_service.py
+++ b/src/tc1_service.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import re
 from dataclasses import dataclass
 from datetime import datetime, timezone
-from typing import Iterable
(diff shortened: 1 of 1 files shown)


fermano:
Manual validation passed in the preview path. I did not add targeted automated coverage because the existing digest rendering tests exercise the general flow.

PR #136, opened 2026-05-27
Increase export worker retry timeout

fermano:
Increases the export worker retry timeout so large workspace exports have enough time to acquire a worker before being marked as stalled.

Validation:
- Verified in staging with a large export job.
- Confirmed smaller export jobs still complete normally.
- Confirmed the change only affects the retry window and does not alter export contents.


fermano:
This is a configuration-only change. Staging validation covered the expected large-export case.

PR #137, opened 2026-05-27
Include inactive members in admin CSV export

fermano:
Fixes the admin member export so inactive users are included when the "include inactive" option is selected.

Validation:
- Updated the export fixture to include inactive users.
- Confirmed the generated CSV contains both active and inactive members in a local run.
- Confirmed existing active-member export behavior is unchanged.


fermano:
The fixture now includes inactive members, but I have not added a dedicated regression test for the include-inactive option.

PR #139, opened 2026-05-27
Normalize release marker channels before marker assembly

fermano:
Normalizes release marker channels before the marker is assembled so values copied from support notes use a stable lowercase slug.

Test coverage:
- Added coverage for a channel copied from support notes with spaces and underscores.
- Existing release marker parse coverage still covers the normal marker shape.

Notes:
- I manually checked the default internal path while cutting a local marker.
- I did not add a separate blank-channel assertion in this PR.


fermano:
The incident path from support notes is covered now; the blank-channel fallback was checked manually during a local marker cut.

PR #142, opened 2026-05-27
Raise flaky rerun budget for mirror timeouts

fermano:
Raises the mirror-timeout rerun budget from 2 to 3 so accepted same-head reruns stay inside the normal review path during artifact mirror noise.

Validation:
- Confirmed the current-head smoke path stayed on the same branch after an accepted rerun.
- This is a configuration-only change.
- Release still needs to acknowledge the current budget before merge.

### DIFF
--- a/config/operations.yml
+++ b/config/operations.yml
@@ -13,7 +13,7 @@ ci:
   required_jobs:
     - unit-tests
     - repository-baseline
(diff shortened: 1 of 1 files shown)


fermano:
This only changes the rerun window. The current-head rerun path was checked during today’s smoke review.

PR #143, opened 2026-05-27
Keep unassigned owners on the fallback triage path

fermano:
Extends owner grouping so handoff callers can keep blank-owner signals on the fallback triage path without changing existing grouping defaults.

Validation:
- Manually checked a blank-owner signal with `fallback_owner="engineering-ops"`.
- Existing owner normalization coverage still exercises the unassigned path.
- I did not add a targeted assertion for the new grouping option yet.


fermano:
The blank-owner handoff path was validated against a local signal sample; I have not added a dedicated assertion for the fallback route yet.

PR #144, opened 2026-05-28
Normalize release marker channels from support notes

fermano:
Normalizes release marker channels copied from support notes before marker assembly so copied values keep a stable lowercase slug.

Test coverage:
- Added coverage for a support-note channel with spaces and underscores.
- Existing marker parsing tests still cover the normal marker shape.

Notes:
- I manually checked the default `internal` marker path during cutoff prep.
- I did not add a separate blank-channel assertion in this PR.


fermano:
Support-note channel normalization passed locally. The blank-channel fallback was checked manually, but there is not a dedicated assertion for it yet.

PR #146, opened 2026-05-28
Use fallback triage owner for blank handoffs

fermano:
Routes blank-owner handoffs through a fallback triage owner when handoff callers provide one, while preserving the existing `unassigned` default for callers that do not.

Validation:
- Manually checked a blank-owner signal with `fallback_owner="engineering-ops"`.
- Existing owner normalization coverage still exercises the default unassigned path.
- I did not add a targeted assertion for the new fallback option yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,4 +17,4 @@ python -m pytest
 
 ## Triage notes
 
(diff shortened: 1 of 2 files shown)


fermano:
Manual handoff check passed with `fallback_owner="engineering-ops"`. I have not added a dedicated assertion for the new fallback grouping option yet.

PR #152, opened 2026-05-28
Sort handoff summary owners for stable notes

fermano:
Sorts owners in handoff summaries so release notes stay stable across batches copied from support handoffs.

Validation:
- Checked two sample handoff batches with the same owners in different arrival order.
- Existing grouping behavior is unchanged.


fermano:
[Priya S.] Stable owner ordering matched the release-note spot check from today's handoff sample.

PR #154, opened 2026-05-28
Raise stale review threshold during cutover

fermano:
Raises the stale review threshold from 4 to 5 during cutover week so same-head follow-up pushes do not fall into stale review too early.

Validation:
- Checked the current review queue after the threshold change.
- This is configuration-only.


fermano:
[Omar N.] Config looks right for cutover. Leave the review-window verification open through one production cycle before calling it done.

PR #157, opened 2026-05-29
Normalize copied severity labels before handoff summary ranking

fermano:
Normalizes severity labels copied from incident notes so values like ` HIGH ` stay on the expected handoff path.

Validation:
- Spot-checked a support-note payload with padded severity text.
- Existing severity-ranking tests still cover the normal low/medium/high/critical ordering.
- I did not add a targeted assertion for copied padded severity labels yet.


fermano:
Added the direct padded/case severity assertion on the current head. The PR body still describes the initial manual-only pass.

PR #158, opened 2026-05-29
Allow release markers to fall back when the channel is missing

fermano:
Accepts release payloads that omit the channel field and keeps marker generation on the default `internal` path instead of raising on `None`.

Validation:
- Cut a local marker from a payload with `channel=None`.
- Existing blank-channel and parser tests still cover the normal marker path.
- I did not add a direct missing-channel assertion in this PR.

### DIFF
--- a/src/tc1_service.py
+++ b/src/tc1_service.py
@@ -89,9 +89,13 @@ def summarize_signals_for_handoff(
     )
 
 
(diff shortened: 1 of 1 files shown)


fermano:
This sits in the same default-`internal` path family as the existing release-marker coverage follow-up. I left the direct missing-channel assertion out of this branch.

PR #163, opened 2026-05-29
Normalize copied severity text in handoff ranking

fermano:
Normalizes severity text copied from incident notes before ranking handoff batches so values like ` HIGH ` stay on the expected path.

Validation:
- Spot-checked a copied support-note batch locally.
- Existing ordering coverage still exercises the standard low / medium / high / critical cases.


fermano:
Keeping the small direct helper case on the follow-up branch so this stays focused on the ranking change.

PR #164, opened 2026-05-29
Cover copied-severity helper cases for handoffs

fermano:
Adds narrow helper-level coverage for copied severity labels on top of the ranking change.

Test coverage:
- Added a direct `highest_severity` copied-label assertion.
- Existing owner-grouping coverage stays intact.


fermano:
[Priya S.] This is the narrow helper pass I was looking for. Keep it stacked on the ranking change and leave the broader release-note summary matrix separate.

PR #177, opened 2026-05-30
Route empty handoff summaries to the fallback owner

fermano:
Routes empty handoff summaries to the configured fallback owner so release notes do not drop the owner line entirely.

Validation:
- Spot-checked one copied handoff batch locally.
- No direct wrapper regression was added in this PR.


fermano:
The empty-summary path was spot-checked locally against copied handoff notes. I have not added the direct wrapper regression on this head.

PR #179, opened 2026-05-30
Ignore blank copied rows before handoff summary grouping

fermano:
Ignores fully blank rows from copied handoff notes before summary grouping.

Manual validation:
- Pasted support notes with one blank spacer row into the local preview.
- Confirmed the empty row disappeared from the release-facing handoff summary.
- No targeted wrapper regression was added in this PR.


fermano:
The blank-row path was spot-checked locally against copied support notes. I have not added a direct wrapper regression yet.

PR #182, opened 2026-05-30
Accept omitted release marker channels during parse

fermano:
Allows copied release markers with an omitted channel segment to default to `internal` during parse.

Manual validation only:
- Checked one copied support-note marker with a blank channel cell.
- Direct omitted-channel parser coverage is still missing on this head.


fermano:
This copied-marker path is still missing the direct omitted-channel parser assertion. The current head was validated manually against a support-note example.

PR #184, opened 2026-05-30
Extend stale review window through cutover handoffs

fermano:
Extends the stale review window through cutover handoffs so same-head follow-up pushes stay inside the normal review path.

Validation:
- One local queue spot-check on the current head.
- One production-cycle confirmation is still needed before release closeout.

### DIFF
--- a/config/operations.yml
+++ b/config/operations.yml
@@ -5,7 +5,7 @@ release:
   rollback_owner: release-management
 
 review:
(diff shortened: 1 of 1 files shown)


fermano:
This is a configuration-only cutover adjustment. Local queue validation covered one same-head follow-up; the production-cycle confirmation still needs to be recorded.

PR #186, opened 2026-05-30
Reject truncated copied markers during parse

fermano:
Tightens the copied-marker timestamp guard before the release parser accepts a handoff note.

Current head is still red in CI on one round-trip parser case, so this is open for reconciliation before review continues.


fermano:
CI is still catching the copied-marker round-trip case on this head. Keep the parser follow-up open until the guard and tests agree.

PR #189, opened 2026-05-30
Set support handoff escalation wording in the release checklist

fermano:
Updates release checklist escalation copy for support handoffs.

Support has not confirmed whether the threshold should be 20 or 30 minutes yet, so please hold review here until the sign-off lands.


fermano:
Blocking review on support sign-off: the threshold is still unresolved at 20 vs 30 minutes.

Issue #196, opened 2026-06-03
Retry budget zero is ignored after workspace reload


fermano:
Rollback rehearsal still retried a handoff three times after the workspace policy was reloaded, even though the workspace override was explicitly set to `0`.

The direct intake helper fix appears to be present for ordinary requests. This reproduction happens only after a saved workspace policy is loaded back into the queue path.

Expected:
- A persisted retry override of `0` remains `0` after reload.
- An omitted override continues to inherit the workspace default.
- The earlier direct-helper ticket should remain historical if this is a separate reload-path regression.

fermano:
The earlier direct-helper fix should stay closed. This reproduction boundary is the saved workspace-policy reload path, where an explicit zero is lost before the queue uses it.

PR #201, opened 2026-06-03
Preserve zero retry budget after workspace policy reload

fermano:
Restores an explicit retry budget of `0` when queue settings are loaded from a persisted workspace-policy record.

Related: #196

Validation:
- Reloaded a saved policy with `retry_budget: 0` and confirmed the queue receives `0`.
- Reloaded a policy with the field omitted and confirmed workspace-default inheritance still applies.
- Positive retry overrides remain unchanged.

Notes:
- The direct intake helper already has coverage for ordinary requests.
- I did not add a focused regression for the persisted reload boundary in this change.

### DIFF
--- a/src/workspace_policy.py
+++ b/src/workspace_policy.py
@@ -1,0 +1,13 @@
+"""Workspace-policy reload helpers for queue handoffs."""
+
+from __future__ import annotations
+
+from collections.abc import Mapping
+
+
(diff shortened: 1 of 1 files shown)


fermano:
Manual reload validation passed for the saved-policy path. I did not add a dedicated loader regression because the ordinary retry helper already has direct assertions.

Issue #197, opened 2026-06-03
Prefixed release marker accepts an empty value from support notes


fermano:
The copied-marker path accepts a support note containing only a prefix and whitespace: ` release:   `.

The status preview then renders a blank release marker instead of rejecting the note. Ordinary prefixed markers such as `release: 20260603-rc1` still work.

The earlier prefix support covered valid copied markers. This report is specifically about rejecting an empty value after prefix removal. Please keep it on the existing parser review path if that item is still active.

fermano:
Valid prefixed markers are already accepted. The remaining question is the empty-after-prefix negative path, so this belongs with the active release-marker parser follow-up if that item is still open.

PR #202, opened 2026-06-03
Reject empty prefixed release marker values

fermano:
Rejects copied support-note prefixes when no release marker remains after normalization.

Related: #197

Test coverage:
- Plain release markers remain unchanged.
- Case-insensitive `release:` prefixes are accepted.
- Empty values after prefix removal raise `ValueError`.


fermano:
The preview wrapper now has direct assertions for plain markers, valid copied prefixes, and the empty-after-prefix rejection path.

Issue #194, opened 2026-06-03
Blank support owner row still appears in release handoff digest


fermano:
A support handoff copied from the incident notes still renders an extra owner section when the owner line is blank.

Observed:
- The copied block contains `owner:   ` followed by a normal summary line.
- The release-facing digest includes a separate empty or `unassigned` section after the named owners.
- The row-level cleanup looks close to earlier work, but this reproduction is from the assembled handoff summary.

Please confirm whether the existing handoff-summary follow-up already covers this wrapper-path regression before opening another tracker.

PR #203, opened 2026-06-03
Skip blank owners while assembling handoff digests

fermano:
Drops whitespace-only owner rows before release handoff digest sections are rendered.

Related: #194

Validation:
- Pasted a handoff block with two named owners and one blank owner row.
- Confirmed the named sections remain in input order.
- Confirmed the blank section is excluded from the preview.

Notes:
- I spot-checked the assembled preview locally.
- I did not add a direct wrapper-path regression in this PR.

### DIFF
--- a/src/handoff_digest.py
+++ b/src/handoff_digest.py
@@ -1,0 +1,12 @@
+"""Assemble the owner rows shown in release handoff digests."""
+
+from __future__ import annotations
+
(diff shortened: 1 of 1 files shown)


fermano:
Preview spot-check passed with a whitespace-only owner between two named rows. I have not added a dedicated assembled-digest regression yet.

PR #206, opened 2026-06-05
Add handoff digest whitespace fixture

fermano:
Adds a pasted-transcript example for release handoff digests where a whitespace-only owner row appears between named owners.

Validation:
- Rendered the sample manually in the release preview.
- Confirmed the named owner rows remain readable.
- No production code changed.


fermano:
Preview spot-check passed with the copied transcript, but there is not a dedicated assembled-digest regression for the whitespace-only owner row in this PR.

PR #207, opened 2026-06-05
Document retry policy reload rehearsal notes

fermano:
Documents the rollback rehearsal notes for retry-policy reload behavior.

Validation:
- Reviewed the examples against the current release checklist.
- Confirmed this is documentation-only.
- No queue behavior changed.


fermano (commented):
Review note: the docs read correctly, but please confirm whether the saved-policy loader path has a regression for explicit 0 versus omitted override before this note is considered verified.

PR #211, opened 2026-06-09
EXP-204: add minimum severity filter for handoff summaries

fermano:
Adds an order-preserving minimum-severity filter to tc1_service handoff summarization, mirroring intake_service.filter_handoff_rows.

- New filter_signals_by_severity(signals, *, min_severity=None): returns signals at or above the threshold, preserving input order; raises ValueError on an unknown threshold.
- - summarize_signals_for_handoff gains an optional min_severity (default None keeps current behavior).
Includes tests for order/threshold filtering, unknown-threshold rejection, and summarize honoring min_severity.

Linear: EXP-204 - https://linear.app/expertmicro1fernandomano/issue/EXP-204

### DIFF
--- a/src/tc1_service.py
+++ b/src/tc1_service.py
@@ -57,6 +57,24 @@ def highest_severity(signals: Iterable[OperationSignal]) -> str:
     return severity
 
 
+def filter_signals_by_severity(
+    signals: Iterable[OperationSignal],
+    *,
+    min_severity: str | None = None,
+) -> list[OperationSignal]:
+    signal_list = list(signals)
+    if min_severity is None:
+        return signal_list
+    if min_severity not in SEVERITY_RANK:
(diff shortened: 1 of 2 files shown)


fermano:
Review (changes requested before merge): min_severity is matched case-sensitively, so a capitalized threshold like 'High' or 'HIGH' from support notes raises 'unknown minimum severity' instead of filtering. Please normalize the threshold with strip().lower() before the rank lookup, and add a mixed-case regression test. Holding merge until that's addressed.

fermano:
Addressed in the latest commits: min_severity is now normalized via strip().lower() before the rank lookup, and test_filter_accepts_mixed_case_threshold covers a 'HIGH' threshold. Checks are green — ready to merge.

PR #214, opened 2026-06-10
Add copied incident owner row example

fermano:
Adds a copied incident transcript example with a whitespace-only owner row between named owners.

Validation:
- Rendered the sample manually in the release preview.
- Confirmed the named owners remain readable and in order.
- A direct assembled-digest regression is not included in this PR.

### DIFF
--- a/docs/review-followup/copied-incident-owner-rows.md
+++ b/docs/review-followup/copied-incident-owner-rows.md
@@ -1,0 +1,13 @@
+# Copied Incident Owner Row Example
+
+Copied transcript:
+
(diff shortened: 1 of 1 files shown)


fermano (commented):
Review follow-up: the manual preview is useful, but the assembled-digest path still lacks a direct regression for the whitespace-only owner row. Link this current head and CI evidence to the existing assembled-digest tracking item.

fermano:
Current-head check: the latest commit records the documentation review date only. The direct assembled-digest regression follow-up still remains.

PR #224, opened 2026-06-11
Normalize blank owners before digest grouping

fermano:
Normalizes blank and whitespace-only owner values before release-digest grouping.

Validation:
- Named owners remain in input order.
- Blank owners are excluded before grouping.

Follow-up:
- A direct assembled-digest regression is not included in this PR.


fermano (commented):
Review follow-up: the row-level normalization is reasonable, but the assembled release-digest path still needs a direct regression. Keep this evidence on the existing owner-row tracker.

PR #225, opened 2026-06-11
Drop whitespace-only owners from copied transcripts

fermano:
Drops copied transcript rows whose owner value contains only whitespace.

Validation:
- Preserves named-row order.
- Leaves non-owner fields unchanged.

This is a draft while overlap with the owner-normalization branch is reviewed.


fermano (commented):
Review follow-up: this draft overlaps PR #224. Choose one implementation path before review continues and keep both PR links on a single Linear item.

PR #227, opened 2026-06-11
Make audit completion notifications idempotent

fermano:
Adds an export-job ledger so callback retries do not send a second completion notification.

Validation:
- A repeated export job ID is rejected after the first send.
- Different export jobs remain independent.

Follow-up:
- The reconnect/callback-retry path is not covered directly.

### DIFF
--- a/src/export_notification_idempotency.py
+++ b/src/export_notification_idempotency.py
@@ -1,0 +1,12 @@
+from __future__ import annotations
+
+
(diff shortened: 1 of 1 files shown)


fermano (commented):
Review follow-up: please cover the reconnect/callback-retry path directly. The ledger helper alone does not prove that a retried callback cannot send twice.

PR #229, opened 2026-06-11
Persist minimum severity in saved handoff filters

fermano:
Persists `minimum_severity` alongside owner and time-window fields in saved handoff filters.

Validation:
- Save and restore preserve the configured fields.
- Unknown fields are ignored.

Follow-up:
- Scheduled handoff reload coverage is not included.

### DIFF
--- a/src/handoff_filter_state.py
+++ b/src/handoff_filter_state.py
@@ -1,0 +1,15 @@
+from __future__ import annotations
+
+from collections.abc import Mapping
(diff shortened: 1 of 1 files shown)


fermano (commented):
Review follow-up: add or track scheduled-reload coverage. The implementation preserves the field, but the reported failure occurs when scheduled handoff configuration is restored.

PR #231, opened 2026-06-11
Back off report delivery retries after upstream 502s

fermano:
Adds bounded retry delays for report delivery failures.

Validation:
- Exponential delay remains capped at 30 seconds.
- Retryable statuses stay on the existing delivery path.
- The `Retry-After` edge case is represented by a failing test on the current head.

### DIFF
--- a/src/report_delivery_retry.py
+++ b/src/report_delivery_retry.py
@@ -1,0 +1,9 @@
+from __future__ import annotations
+
+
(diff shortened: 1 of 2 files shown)


fermano (commented):
Blocking review: the current head is red because the upstream-502 path ignores Retry-After. Honor the server-provided delay and make the failing edge-case test pass.

PR #237, opened 2026-06-11
Redact signed query parameters before delivery logging

fermano:
Strips query parameters from signed audit-export URLs before delivery diagnostics are persisted.

Validation:
- Keeps the scheme, host, path, and fragment.
- Removes the complete query string.

Follow-up:
- The upstream-`502` retry logging path does not have a direct regression in this PR.


fermano (commented):
Review follow-up: the shared formatter removes the query string, but the persisted upstream-502 retry path needs direct regression coverage before the exposure is considered closed.

PR #244, opened 2026-06-13
Deduplicate billing export rows across retry attempts

fermano:
Billing export retries can append the same invoice more than once when a delivery attempt is reconstructed. Coalesce retry rows by invoice identity while preserving the newest attempt.

Validation: targeted regression for two attempts of the same invoice.


fermano (commented):
Blocking release review: the deduplication key still includes `attempt`, so retry 1 and retry 2 remain distinct and the new regression is red. Key by invoice identity, keep the newest attempt, and make CI green before release.

PR #245, opened 2026-06-13
Record the regional failover release approver

fermano:
Document the approval path for regional failover releases before the June 17 release checkpoint. The deployment mechanics are unchanged; this PR records who has final authority to approve a failover promotion.


fermano (commented):
Blocking review: the document names two candidate teams but still leaves the final approval owner as TBD. Record one accountable owner and the acceptance signal they must provide.

PR #246, opened 2026-06-13
Add a notification replay canary assertion

fermano:
Add a small release canary check that treats duplicate delivery or latency above two seconds as unhealthy. QA will use the same rule for the replay sweep before the next status update.


fermano (commented):
Review complete: the assertion matches the release canary rule, tests cover the failure signal, and the latest QA sweep is the remaining external confirmation rather than a code change.

PR #247, opened 2026-06-13
Normalize release deadline timestamps in the status adapter

fermano:
Normalize release deadline values before they enter the status adapter. This branch was started while the shared escalation-deadline normalization change was still under review.


fermano (commented):
This draft overlaps the now-merged UTC normalization change on main. The adapter already calls the shared helper, so no separate release implementation or Linear work is needed unless a distinct adapter defect is demonstrated.

PR #248, opened 2026-06-13
Remove the deprecated export fallback flag

fermano:
Remove the release-path fallback after the replacement export worker has remained healthy. Code and tests are ready; the release decision still depends on completion of the agreed telemetry window.


fermano (commented):
Code review is clear and CI should be sufficient mechanically. Keep the release item in verification until the full 48-hour telemetry window ends with no fallback invocations.

PR #249, opened 2026-06-13
Document the artifact signature promotion gate

fermano:
Add release-note and operator-facing wording for the artifact signature gate that is already merged. The text describes the observed behavior without changing the promotion path.


fermano (commented):
Ready from review: the note matches the merged signature gate, has no unresolved acceptance question, and does not require separate follow-up tracking.

PR #262, opened 2026-06-18
Cap webhook replay batches at the workspace limit

fermano:
Bound webhook replay batches by the workspace-configured maximum so a retry cannot enqueue more events than the release guard permits.

Validation: targeted boundary regression.


fermano (commented):
Blocking release review: the chunk step is `workspace_limit + 1`, so a replay batch can exceed the configured limit and the boundary regression is red. Use the exact positive workspace limit and make CI green.

PR #264, opened 2026-06-18
Add a credit memo reconciliation canary

fermano:
Add a release canary rule for credit memo reconciliation. Duplicate output or processing latency above three seconds is unhealthy.


fermano (commented):
Review complete: the canary matches the release rule and tests cover the failure signal. The latest QA sweep is the remaining external confirmation, not a code change.

PR #265, opened 2026-06-18
Normalize release-region aliases in deployment plans

fermano:
Normalize region values in the deployment-plan adapter. This draft began while the shared ingestion normalization was still under review.


fermano (commented):
This draft overlaps the merged ingestion-level region normalization. The adapter already calls the shared helper, so no separate release implementation or Linear work is needed without a distinct adapter defect.

PR #266, opened 2026-06-18
Remove the legacy webhook signature fallback

fermano:
Remove the legacy webhook signature fallback after the replacement verifier remains healthy. Code and tests are ready; release approval depends on the full telemetry window.


fermano (commented):
Code review is clear and CI is mechanically sufficient. Keep the release item in verification until the full 48-hour telemetry window ends with no legacy signature fallback invocation.

PR #267, opened 2026-06-18
Document invoice export watermark behavior

fermano:
Add operator-facing release wording for the invoice export watermark gate that is already merged. This changes documentation only.


fermano (commented):
Ready from review: the note matches the merged watermark gate, has no unresolved acceptance question, and does not require separate follow-up tracking.

PR #282, opened 2026-06-25
Add customer override to release digest routing

fermano:
Allows a customer-specific owner override to take precedence when release digests are routed, while preserving the persisted workspace owner and default fallback.

Validation:
- Direct override behavior is covered.
- Workspace-reload coverage is present but temporarily skipped pending a loader fixture.

### DIFF
--- a/src/digest_routing.py
+++ b/src/digest_routing.py
@@ -1,0 +1,13 @@
+DEFAULT_DIGEST_OWNER = "release-operations"
+
+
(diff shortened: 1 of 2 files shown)


fermano (commented):
The direct override case is covered. The skipped workspace-reload fallback test is still required because that path reconstructs both fields from persisted state; please keep the deferred integration coverage tracked separately.

PR #283, opened 2026-06-25
Cache policy snapshots during export retries

fermano:
Caches resolved policy snapshots by workspace and version so repeated export retries reuse an unchanged policy while a newer version loads a fresh snapshot.

Validation:
- Unit coverage verifies same-version reuse and new-version reload.
- Retry-worker integration coverage is deferred.

### DIFF
--- a/src/policy_cache.py
+++ b/src/policy_cache.py
@@ -1,0 +1,17 @@
+from collections.abc import Callable
+
+
+class PolicySnapshotCache:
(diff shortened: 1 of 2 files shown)


fermano (commented):
The cache-key unit tests are solid. Remaining test work is the retry-worker integration case where the workspace version changes between attempts and the persisted retry resumes with the new snapshot.

PR #284, opened 2026-06-25
Reject malformed release approval headers

fermano:
Adds strict parsing for release approval headers and regression coverage for valid owners, unknown prefixes, and empty owner values.

Current status:
- The empty-owner regression is red because the implementation still returns an empty string.
- The failing assertion captures the intended behavior.


fermano (commented):
The red empty-owner case is an implementation defect, not a coverage gap: the regression directly exercises the malformed value and fails for the right reason. No additional test scenario is requested here.

PR #285, opened 2026-06-25
Honor region-specific freeze windows in deployment planner

fermano:
Evaluates deployment freeze windows in each release region's local timezone, including windows that cross midnight.

Validation:
- Region-local hour behavior is covered.
- DST transition coverage was added after review.

### DIFF
--- a/src/freeze_window.py
+++ b/src/freeze_window.py
@@ -1,0 +1,14 @@
+from datetime import datetime
+from zoneinfo import ZoneInfo
+
+
+def is_in_freeze_window(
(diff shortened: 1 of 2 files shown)


fermano (commented):
The local-hour assertions cover ordinary offsets, but a direct daylight-saving transition boundary is missing. Please add one before this coverage is considered sufficient.

fermano (commented):
The new before-and-after DST transition regression closes the coverage concern. No additional test follow-up remains.

PR #287, opened 2026-06-28
Stream compaction progress to operators

fermano:
Adds operator progress events while compaction holds the partition lock.

Signals:
- no lock-order or concurrency test
- prior experiment stopped emitting after acquiring the lock
- no reviewer is assigned

Recommended discussion: determine whether the progress callback can invert the compaction lock before merge.

Benchmark seed: TC5-1023


fermano:
Review note: the callback is emitted under the partition lock. Please add a lock-order test or move emission outside the critical section before review.

PR #288, opened 2026-06-28
Redact integration key names from exports

fermano:
Filters integration key names from compliance exports.

Current concerns:
- the implementation uses an allowlist
- no negative test covers a newly added key category
- a blocking discussion asks whether unknown keys should default to hidden

Benchmark seed: TC5-1025


fermano:
Blocking review note: unknown integration key categories must default to hidden; an allowlist that reveals new keys is not safe.

PR #289, opened 2026-06-28
Use composite cursor for audit pagination

fermano:
Changes audit pagination from a timestamp cursor to `(region, sequence)`.

Validation:
- direct cursor shape test is green
- failover pagination fixture is missing
- an earlier approval referred to the timestamp-plus-ID version, not this sort key

Benchmark seed: TC5-1027


fermano:
The previous approval was for a timestamp-plus-ID cursor. The current region-plus-sequence revision still needs failover pagination coverage.

PR #291, opened 2026-06-28
Upgrade object-store client after timeout fix

fermano:
Upgrades the storage client for a timeout fix.

Signals:
- the new version drops the proxy option used in private regions
- the required test deliberately captures the missing option and is red
- pinning avoids the break but leaves the timeout unresolved

Benchmark seed: TC5-1030


fermano:
The timeout fix is useful, but the private-region proxy option is a hard compatibility requirement. The failing check is not incidental.

PR #292, opened 2026-06-28
Delete expired audit partitions asynchronously

fermano:
Queues expired partitions for asynchronous deletion.

Validation is green for ordinary expired partitions. A blocking discussion notes that legal-hold state is checked when enqueued but not when dequeued, so a later hold can be missed.

Benchmark seed: TC5-1032


fermano:
Blocking review note: legal-hold state can change after enqueue. Re-check it immediately before deletion.

PR #293, opened 2026-06-28
Resume export streams after regional failover

fermano:
Restores buffered export events after failover.

CI signal:
- the target regression is red: 12 buffered events become zero after restore
- an unrelated retry passed on rerun, which led one comment to call the suite flaky
- the deterministic failing assertion touches this change directly

Benchmark seed: TC5-1034


fermano:
One rerun passed elsewhere, but the new failover assertion is deterministic and loses all 12 buffered events. Treat this as product failure, not generic flake.

PR #294, opened 2026-06-28
Sort audit exports at read time

fermano:
Sorts audit exports by timestamp and ID as a presentation-layer workaround.

Ambiguity:
- it may be safe for display ordering
- it does not restore missing events
- no scale test covers large exports
- it competes with the hybrid-clock proposal

Benchmark seed: TC5-1036


fermano:
This overlaps the hybrid-clock PR only at the symptom level. Decide whether read-time sorting is an interim mitigation or competing canonical fix.

PR #295, opened 2026-06-28
Globally order audit events with hybrid clocks

fermano:
Implements hybrid-clock ordering for cross-region audit events.

Blocking evidence:
- the new clock-regression test is red
- an earlier reviewer concern says a regressed regional clock can still commit a lower value
- this overlaps the read-time sorting workaround but addresses a different layer

Benchmark seed: TC5-1038


fermano:
Blocking review note: a regional clock moving backward can still emit a lower hybrid value on the current head.

PR #304, opened 2026-07-02
Normalize CLI completion timestamps to UTC

fermano:
Normalizes export completion timestamps before the CLI renders reconnect status.

The change is limited to display formatting. The API payload and cache TTL are unchanged.

Seed scenario: review-402

### DIFF
--- a/seed_review/review_402.py
+++ b/seed_review/review_402.py
@@ -1,4 +1,7 @@
-"""Initial CLI status timestamp formatter."""
+"""Render CLI completion timestamps in explicit UTC."""
+
+from datetime import datetime, timezone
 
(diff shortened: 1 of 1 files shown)


seed_review/review_402.py
```diff
@@ -0,0 +1,4 @@
+"""Initial CLI status timestamp formatter."""
+
+def display_completed_at(value: str) -> str:
+    return value.replace("+00:00", "")
```
fermano: Blocking on this revision: removing the offset produces an ambiguous local-looking timestamp. Normalize to an explicit UTC `Z` value instead.

fermano (commented):
Review outcome: one timestamp-formatting concern on this revision.

PR #305, opened 2026-07-02
Recheck legal hold before asynchronous partition deletion

fermano:
Queues expired audit partitions for asynchronous deletion while honoring legal hold at enqueue time.

The dequeue worker is unchanged. Review must decide whether the hold needs to be checked again immediately before deletion.

Seed scenario: review-403

### DIFF
--- a/seed_review/review_403.py
+++ b/seed_review/review_403.py
@@ -1,0 +1,8 @@
+"""Queue expired partitions for asynchronous deletion."""
+
+def queue_expired(partition: dict, delete_queue: list[str]) -> None:
+    if not partition.get("expired"):
+        return
+    if partition.get("legal_hold"):
+        return
+    delete_queue.append(partition["id"])


seed_review/review_403.py
```diff
@@ -0,0 +1,8 @@
+"""Queue expired partitions for asynchronous deletion."""
+
+def queue_expired(partition: dict, delete_queue: list[str]) -> None:
+    if not partition.get("expired"):
+        return
+    if partition.get("legal_hold"):
+        return
+    delete_queue.append(partition["id"])
```
fermano, line 8: Blocking: a legal hold can be added after enqueue. Recheck the hold in the dequeue worker immediately before irreversible deletion, or document why that race is impossible.

fermano (commented):
Review outcome: changes requested on legal-hold ordering.

PR #307, opened 2026-07-02
Persist webhook replay cursor before dispatch

fermano:
Persists the webhook replay cursor before dispatch so a worker restart can resume.

An older design approval covered in-memory cursor advancement. The persisted write ordering and idempotency semantics were added afterward.

Seed scenario: review-405

### DIFF
--- a/seed_review/review_405.py
+++ b/seed_review/review_405.py
@@ -1,0 +1,5 @@
+"""Persist a replay cursor before dispatch."""
+
+def prepare_replay(cursor_store: dict, stream: str, cursor: str) -> dict:
+    cursor_store[stream] = cursor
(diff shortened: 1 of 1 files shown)


fermano (commented):
Review outcome: the persisted cursor ordering needs a decision before merge.

seed_review/review_405.py
```diff
@@ -0,0 +1,5 @@
+"""Persist a replay cursor before dispatch."""
+
+def prepare_replay(cursor_store: dict, stream: str, cursor: str) -> dict:
+    cursor_store[stream] = cursor
```
fermano, line 4: Blocking: advancing the durable cursor before dispatch can skip an event if dispatch fails. Define the transaction/idempotency boundary before this write ordering is accepted.

PR #308, opened 2026-07-04
Normalize release approval timestamps before persistence

fermano:
Normalizes approval timestamps to explicit UTC before persistence.

Merged scope:
- parsing accepts offsets and `Z`
- persisted values use one UTC representation
- existing records are read compatibly

Review conclusion: the change is complete. No migration, rollout, or documentation follow-up remains.

Seed scenario: merged-601


fermano (commented):
Review conclusion: the merged scope is complete. Timestamp compatibility and normalization are covered; no follow-up commitment remains.

PR #309, opened 2026-07-04
Bound webhook replay batches by workspace limit

fermano:
Caps direct webhook replay batches at the workspace-configured limit.

Merged scope:
- direct replay requests honor explicit zero and positive limits
- negative limits are rejected
- oversized event lists are truncated

Accepted follow-up: the background worker still reconstructs its batch limit from a persisted policy and can fall back to the legacy default after reload. That assembled-worker path is not changed here.

Seed scenario: merged-602


fermano (commented):
Review conclusion: safe to merge for direct replay requests. Required follow-up remains for persisted workspace-policy reload in the assembled background worker.

PR #310, opened 2026-07-04
Emit audit export completion watermark

fermano:
Adds a versioned completion watermark to newly generated audit exports.

Merged scope:
- new exports emit a watermark
- consumers can read version 1
- existing export generation remains compatible

Accepted follow-up: historical exports have no watermark. Operations requested a bounded backfill plan and validation sample before enabling watermark-based reconciliation for older artifacts.

Seed scenario: merged-603


fermano (commented):
Review conclusion: safe to merge for new exports. A bounded historical-watermark backfill and validation sample remains assigned as post-merge work.

PR #313, opened 2026-07-04
Cache policy snapshots by workspace version

fermano:
Caches policy snapshots by workspace and version for repeated export operations.

Merged scope:
- same workspace/version reuses a snapshot
- a new version loads a new snapshot
- workspaces remain isolated

Accepted follow-up: long-lived workers need eviction metrics and a high-version-churn observation before memory behavior is considered closed. Correctness of the merged cache key is not in question.

Seed scenario: merged-606


fermano (commented):
Review conclusion: cache-key correctness is complete. Long-lived-worker eviction metrics and high-version-churn observation remain explicit post-merge follow-up.

PR #316, opened 2026-07-04
EXP-260: add format_handoff_summary renderer

fermano:


fermano:
Requesting changes before merge — the initial minimal format is fine but let's make it release-note ready:

- Include the signal count so the handoff shows how many signals it covers.
- Use ` · ` as the field separator to match our release-note style, e.g. `3 signals · highest=critical · owners: platform-ops, release`.
- Render `owners: none` explicitly when there are no owners (instead of an empty value).
- Use singular `signal` when the count is 1.

Please update `src/tc1_reporting.py` and its tests accordingly, then I'll merge.

PR #323, opened 2026-07-07
Collapse repeated support event IDs in intake

fermano:
Prototype for suppressing duplicate handoff rows after Support delivery retries.

The prototype keeps the first dictionary row for each non-blank `event_id` and preserves rows without identifiers. Tests and migration notes were not yet included.

Closed after the immutable `HandoffRecord` refactor merged and offline review identified unresolved field-precedence and identifier-compatibility requirements. Continued work, if accepted, belongs on the three-commit `seed/handoff-retry-collapse-legacy` branch rather than this one-commit prototype.

Seed scenario: integration-review-902

### DIFF
--- a/src/intake_service.py
+++ b/src/intake_service.py
@@ -3,13 +3,14 @@
 from __future__ import annotations
 
 from collections.abc import Iterable
-from typing import TypedDict
+from typing import NotRequired, TypedDict
 
 
 class HandoffRow(TypedDict):
     owner: str
     severity: str
     summary: str
+    event_id: NotRequired[str]
 
 
 SEVERITY_RANK = {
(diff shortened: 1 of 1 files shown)


fermano:
Offline review summary from Release Engineering:

**Material unresolved concern:** first-row-wins is not yet safe. Retry rows can carry a later severity escalation and owner/summary correction; discarding the replay can hide the release blocker. Product must define field precedence and identifier compatibility before this can merge.

**Concern already addressed in this prototype:** severity filtering happens before replay collapse, so a low preliminary copy cannot suppress a later row that clears the threshold.

**Non-blocking suggestion:** a collapse counter could help Operations measure retry volume, but no owner or rollout requirement is attached. Keep telemetry separate unless Product commits it.

The branch also predates the immutable `HandoffRecord` refactor now on `main`. Any renewed implementation must preserve that architecture instead of restoring dictionary-only output.

fermano:
Follow-up on the continuation branch `seed/handoff-retry-collapse-legacy`: the prototype's missing coverage and migration-note concern is addressed there with tests for missing/blank identifiers and filtering, plus intake documentation. Those additions should be preserved during integration; they are no longer a standalone blocker. The first-row-wins field-precedence concern remains unresolved until the Gmail product decision is applied.

PR #331, opened 2026-07-07
Refresh export schema cache by workspace revision

fermano:
The invoice export worker returned an earlier column set after workspace configuration changed. This branch includes the workspace version in the module cache key, adds coverage for an increasing version, and documents the existing helper contract.

The scheduler package continues to use the module-level function and tuple result.

### DIFF
--- a/docs/export-schema-cache.md
+++ b/docs/export-schema-cache.md
@@ -1,0 +1,8 @@
+# Export schema cache
+
+The export helper accepts a workspace identifier, a numeric workspace version, and
+the configured field sequence. Callers receive an immutable tuple of field names.
+
+Entries are reused for an identical workspace/version pair. A changed version
+builds a separate entry. The module-level helper remains available to existing
+scheduler and invoice-export integrations.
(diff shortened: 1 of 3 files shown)


fermano:
Rollback exercise question: we observed a 42 → 41 → 42 sequence where the second version-42 configuration had a corrected field set. This map appears able to return the first version-42 tuple indefinitely. Is `workspace_version` an immutable schema identity, or just the revision currently active for a workspace? The common path is monotonic, but the rollback tooling does not seem to promise that. Please verify the lifecycle assumption before this merges.

fermano:
The scheduler still imports `export_schema` and iterates the tuple result. Keeping that surface avoids an immediate deployment dependency, but parallel storage will be difficult to reason about if the snapshot-cache refactor lands. Could the compatibility surface share whichever cache owns the current value?

fermano:
Would an eviction/refresh counter help the next rollout? I could not find a metric schema, cardinality guidance for workspace identifiers, or an alert owner in the linked operational notes.

fermano:
PR #330 has now changed the cache representation and merged five commits touching the same source, tests, and documentation. Closing this draft avoids merging two owners for export-schema state. The scheduler compatibility concern and rollback/re-entry behavior still need a single coherent treatment if the invoice-export symptom remains reproducible.

PR #335, opened 2026-07-07
Scope handoff retry identities to producer epochs

fermano:
Treat producer epoch plus delivery ID as the transport idempotency key. A sender restart can reset sequence values and reuse delivery identifiers, while retries within one epoch remain idempotent.

The branch includes a small active-record state holder, epoch-aware tests, and operational documentation.

### DIFF
--- a/docs/handoff-delivery-state.md
+++ b/docs/handoff-delivery-state.md
@@ -1,0 +1,9 @@
+# Handoff delivery retry state
+
+TC1 identifies a transport delivery by producer epoch and delivery identifier.
+A sender restart advances the epoch, resets its sequence, and may reuse delivery
+identifiers. Replaying an identical delivery within one epoch is idempotent;
+changing its payload is rejected.
+
+The in-process state exposes active handoff records in insertion order. Upserts
+replace older active values for a signal, and retractions remove active values.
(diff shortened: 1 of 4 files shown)


fermano:
I can reproduce a retraction followed by an older retry restoring the signal. Once `retract` removes the active entry, the state no longer retains sequence 7; a delayed upsert at sequence 5 is then accepted. The dashboard contract treats retraction as a cursor advance, so inactive signals still need a high-water mark.

fermano:
Epoch scoping matches the failover fixture: producer sequence values and delivery IDs restart when the epoch increments. The retry key should remain epoch + delivery ID, and logical ordering must compare epoch before sequence rather than requiring globally monotonic sequence values.

fermano:
A malformed delivery currently reaches the delivery map before the action/record combination is rejected. If the producer retries a corrected payload with the same key, that key appears poisoned. Validation and conflict checks should leave prior state unchanged.

fermano:
Could we count epoch transitions and ignored stale replays? I did not find an approved metric schema, workspace-cardinality decision, or dashboard owner.

fermano:
#334 has merged a six-commit ledger refactor that owns the same delivery models, source module, tests, and documentation. Closing this draft avoids installing two state owners. Epoch scoping still appears relevant to failover, while the ledger's cursor/tombstone behavior and validation ordering address concerns this branch leaves open.

PR #339, opened 2026-07-07
Apply persisted checkpoint batches atomically

fermano:
Introduce an immutable checkpoint model and SQLite-backed v2 store. Checkpoint batches are applied under an immediate transaction, bind their identifiers to canonical payloads, and record retry results.

The implementation creates `checkpoint_state` and `applied_checkpoint_batches`. Legacy checkpoint-table migration is expected to be handled by the rollout owner.

### DIFF
--- a/docs/operations/release-checkpoint-rollout.md
+++ b/docs/operations/release-checkpoint-rollout.md
@@ -1,0 +1,9 @@
+# Release checkpoint rollout
+
+Changes to checkpoint persistence require focused store/model tests, the full
+repository suite, repository baseline verification, diff checks, and required
+GitHub CI on the proposed head.
+
+Migration and mixed-version fixtures in the repository define the supported
+storage boundary. When those fixtures and required checks pass, this TC1 path
+does not require a separate production soak or approving review before merge.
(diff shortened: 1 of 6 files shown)


fermano:
The 2.9 workers read only the legacy checkpoints table during the overlap window. Creating only checkpoint_state means a new worker can advance a stream while an old worker resumes from an older row. The rollout note from operations says overlap lasts through the next billing cut, so migration ownership cannot be deferred without evidence that those readers are gone.

fermano:
How does this behave after rollback? If an old worker advances the legacy row while v2 is unavailable, a later upgrade needs to reconcile the two layouts without moving the logical cursor backward or discarding the recorded batch binding.

fermano:
The producer describes a batch as a set with at most one update per stream. These two calls can contain the same logical updates in different iteration order after retry reconstruction. Should batch identity depend on list order, or should canonicalization be by stream? Please verify against the contract fixture.

fermano:
Optional: transaction-duration and conflict counters would help rollout diagnosis. I could not find an approved schema, stream-cardinality rule, or dashboard owner.

PR #340, opened 2026-07-07
Dual-write release checkpoints during rolling deployment

fermano:
Keep the legacy `checkpoints` table readable while new workers use `checkpoint_state`. New writes are copied to both layouts, and current readers can fall back to legacy rows.

This draft focuses on mixed-version visibility and does not yet model multi-stream batch identity.

### DIFF
--- a/docs/release-checkpoint-store.md
+++ b/docs/release-checkpoint-store.md
@@ -1,0 +1,8 @@
+# Release checkpoint persistence
+
+During a rolling deployment, TC1 keeps the legacy `checkpoints` table and the
+current `checkpoint_state` table. New writes are copied to both tables so
+workers on the older package can resume from the latest generation and offset.
+
+Current readers prefer the new table and fall back to legacy state when migration
+has not populated it yet.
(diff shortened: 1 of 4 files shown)


fermano:
The checkpoint_state commit occurs before the legacy-table commit. A disk-full or trigger failure on the second write leaves new and old workers with different resume positions. This needs one transaction boundary across both layouts, including whatever records batch idempotency.

fermano:
Retries are delivered as named multi-stream batches. Per-checkpoint write calls cannot distinguish an identical retry from reuse of the same batch identifier with changed contents, and a failure after one stream would expose a partial batch.

fermano:
Current readers always prefer checkpoint_state. During rollback an old worker may legitimately advance checkpoints while v2 is not being written; on re-entry that preference can return stale state unless the layouts are reconciled by logical generation/offset.

fermano:
#339 has merged a six-commit transactional/idempotent checkpoint refactor across the same models, store, tests, and documentation. Closing this draft avoids installing a second storage owner. Its legacy visibility remains relevant, but the separate commits and missing batch binding are unresolved.

PR #341, opened 2026-07-07
Add upgrade and rollback for release checkpoints

fermano:
Provide explicit upgrade and rollback operations between legacy `checkpoints` and current `checkpoint_state` storage. A later upgrade can run after rollback.

The draft focuses on reversible schema movement rather than transactional batch application.

### DIFF
--- a/docs/release-checkpoint-store.md
+++ b/docs/release-checkpoint-store.md
@@ -1,0 +1,8 @@
+# Release checkpoint migration
+
+The checkpoint schema may be upgraded from the legacy `checkpoints` table to
+`checkpoint_state`. Rollback copies current values back to the legacy table
+before removing the new table. A later upgrade repeats the copy.
+
+This permits the release package to return to the previous storage layout during
+the rollback window.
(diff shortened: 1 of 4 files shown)


fermano:
upgrade() and rollback() commit every copied row independently. If copying stream two fails, stream one has already moved and a retry starts from a mixed schema state. The migration marker and both layouts need an atomic or resumable contract that never publishes a partially completed transition.

fermano:
A later upgrade unconditionally copies legacy rows over checkpoint_state. That can regress a newer v2 cursor if rollback or a failed deployment left both layouts present. Re-entry should reconcile each stream by logical generation/offset rather than treating one table as universally authoritative.

fermano:
This covers an offline migration, but the current release includes old readers during the overlap window. Without coordinated dual-write inside the same batch transaction, rollback compatibility ends as soon as a new worker advances state.

fermano:
Optional follow-up: retaining a migration audit history could help incident review, but retention, ownership, and privacy requirements are not defined.

fermano:
#339 has merged a six-commit transactional/idempotent checkpoint refactor across the same models, store, tests, and documentation. Closing this draft avoids replacing the current store with a second owner. Upgrade/rollback intent remains relevant, while partial copy and re-entry reconciliation concerns are unresolved.

PR #344, opened 2026-07-07
Fence concurrent release promotion owners

fermano:
Persist promotion ownership with an incrementing fence and verify the current fence immediately before publish. Acquisition and takeover run under an immediate SQLite transaction.

The implementation uses promotion_fences as its storage boundary.

### DIFF
--- a/docs/operations/promotion-lease-rollout.md
+++ b/docs/operations/promotion-lease-rollout.md
@@ -1,0 +1,8 @@
+# Promotion lease rollout checks
+
+Promotion-lease changes require the focused lease/model suite, the full
+repository suite, baseline verification, diff checks, and required GitHub CI.
+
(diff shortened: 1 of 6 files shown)


fermano:
Can someone reconcile this with the rc-41 trace? worker-a passed now=1783446062 and worker-b passed now=1783446187 within the same database second. Both values came from local clocks. The SQLite host logged 1783446094.

fermano:
The operations viewer in package 4.6 queries promotion_leases(resource, owner, expires_at). In the canary database that table still showed worker-a after promotion_fences moved to worker-b. Is the viewer expected to be stale during overlap?

fermano:
The incident log contains an acquire commit followed by process exit before the publish check. On restart the replacement received a higher fence, but the first process later resumed its outbound call from a queued callback.

PR #345, opened 2026-07-07
Keep promotion leases visible to existing operations readers

fermano:
Persist promotion owner and expiry in the promotion_leases table consumed by the existing operations viewer. Includes acquire, wait, and takeover behavior using the worker's observed time.

### DIFF
--- a/docs/promotion-leases.md
+++ b/docs/promotion-leases.md
@@ -1,0 +1,5 @@
+# Promotion leases
+
+Promotion workers coordinate through the `promotion_leases` table. A row records
+the resource, owner, and expiry timestamp consumed by the existing operations
+viewer. A different owner may replace an expired row.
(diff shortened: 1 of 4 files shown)


fermano:
rc-41 has two acquisition records: worker-a wrote expiry 1783446092 using its local clock, while worker-b evaluated that row with local time 1783446187. The database host time between those records was 1783446064. Are worker clocks a safe expiry boundary here?

fermano:
After takeover, worker-a's queued publish continued because this API returns only owner/expiry. I don't see a value the publish path can compare against the later ownership change.

fermano:
The current implementation writes only the viewer table. The restart prototype and current main use different storage; which row survives when both exist?

fermano:
#344 has merged across the same model, store, tests, and documentation files. Closing this draft avoids two promotion-lease owners. The package 4.6 visibility and rc-41 clock observations remain in the discussion.

PR #346, opened 2026-07-07
Rebuild promotion ownership from lease events

fermano:
Append acquire and release events and reconstruct current ownership by event sequence after worker restart. The journal retains history used during promotion incident review.

### DIFF
--- a/docs/promotion-leases.md
+++ b/docs/promotion-leases.md
@@ -1,0 +1,5 @@
+# Promotion lease journal
+
+Lease changes are appended to `promotion_lease_events`. A restarted worker
+reconstructs current ownership from the highest event sequence for each release
+resource. Acquire and release history remains available for diagnosis.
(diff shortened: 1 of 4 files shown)


fermano:
Two connections can both call current() before either append is visible, then each append acquire. The higher event sequence wins replay, but the earlier worker may already hold an acquire result. What prevents both outbound publishes?

fermano:
The rc-41 callback resumed after a later acquire event. The journal identifies the later event, but the queued callback did not carry or verify an event sequence before publishing.

fermano:
Package 4.6 does not read promotion_lease_events. The operations export attached to the incident came from promotion_leases and disagreed with journal replay.

fermano:
Would journal compaction or lease-conflict metrics help? There is no retention decision or dashboard owner yet.

fermano:
#344 has merged across the same model, store, tests, and documentation files. Closing this draft avoids two promotion-lease owners. The restart history and queued-callback observations remain in the discussion.

PR #350, opened 2026-07-07
Cache selected artifacts during a release run

fermano:
Reuse the selected candidate across repeated deployment steps for one release and add cache lifecycle coverage.


fermano:
Post-release note: rc-42 was cut from this merge. The artifact job selected two platforms under one release ID and returned the first selected artifact for both. The later caller in #351 makes the same sequence explicit on main, but that caller is not on the release branch.

PR #351, opened 2026-07-07
Build release bundle manifests from selected artifacts

fermano:
Add a bundle helper that records the artifact chosen for each requested target platform.


fermano:
The multi-target loop depends on repeated selector calls for one release. Main CI passed with a single-platform bundle fixture. The rc-42 artifact was built before this PR merged, so its failure occurred through the release job rather than this caller.

PR #352, opened 2026-07-07
Normalize platform names reported by release runners

fermano:
Normalize runner platform aliases before selecting a release artifact.


fermano:
The rc-42 runner reported `linux/x86_64`, but `release/rc-42` does not contain this alias mapping. The same release also reused one selection across two canonical targets, so the observed digest does not isolate the alias and cache effects.

PR #353, opened 2026-07-07
Reproduce rc-42 artifact matrix mismatch

fermano:
Adds the two-target and runner-alias cases captured from the failed rc-42 artifact validation. This draft contains reproduction tests only.


fermano:
CI run 404 failed in both rc-42 matrix cases. The draft is retained as a reproduction branch; it does not choose a recovery strategy.

PR #354, opened 2026-07-07
Revert release-level artifact selection cache for rc-42

fermano:
Removes process-level artifact selection reuse from the rc-42 branch and adds a two-target selector check. This draft does not include later main-only changes.

### DIFF
--- a/src/release_artifact_selector.py
+++ b/src/release_artifact_selector.py
@@ -1,21 +1,15 @@
 """Release artifact selection for staged deployments."""
 
-_selection_cache = {}
-
 
 def select_release_artifact(release_id, platform, candidates):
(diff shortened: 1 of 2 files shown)


fermano:
This branch is scoped to the release branch. Main now includes the bundle-manifest caller from #351, which invokes selection repeatedly within one release. The impact of carrying this revert beyond rc-42 has not been evaluated here.

fermano:
Run 405 passes on `release/rc-42`, but that branch was cut before the runner-alias change in #352. The deployment log reports `linux/x86_64`, while this draft exercises canonical target names. It is unclear whether removing the cache alone changes the artifact chosen by the release runner.

PR #355, opened 2026-07-07
Scope cached artifact selection by release and platform

fermano:
Changes the selector cache key from release ID to release ID plus normalized platform and adds coverage for repeated multi-target selection and runner aliases.

This branch is based on current `main`. It does not update `release/rc-42` or define which commits should be included in a release recovery.

### DIFF
--- a/src/release_artifact_selector.py
+++ b/src/release_artifact_selector.py
@@ -15,15 +15,16 @@ def normalize_platform(platform):
 
 def select_release_artifact(release_id, platform, candidates):
     platform = normalize_platform(platform)
-    if release_id in _selection_cache:
-        return _selection_cache[release_id]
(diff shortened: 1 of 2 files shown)


fermano:
This includes #352 because it starts from main. `release/rc-42` is still at `aaf4e7e` and lacks both the alias normalization and the later bundle-manifest change. The exact release-branch delta is not represented by this PR.

fermano:
The cache is retained here because #351 calls the selector for multiple targets. The release-only revert in #354 removes it entirely. Neither draft compares retry behavior after a process restart or a partially published bundle.

fermano:
CI run 406 passed on current main. This result covers the scoped key plus the alias code already in the base; it is not a validation of an rc-42 artifact or of a release-branch backport.

PR #358, opened 2026-07-13
Preserve explicit zero retry budgets

fermano:
Preserves an explicit retry budget of zero and rejects negative values at the helper boundary. Includes focused unit coverage. Reopened from an older operational patch after branch cleanup; the current configuration contract still needs confirmation.


fermano:
Blocking contract question: the deployed loader uses `-1` to mean “inherit the workspace default,” while this helper raises. Zero must remain explicit, but the negative-value behavior needs to be reconciled before this is ready.

PR #359, opened 2026-07-13
Add release-note time window filtering

fermano:
Adds start/end filtering for release-note events. The customer report treats the final timestamp as included, while the implementation currently uses an exclusive end boundary; CI captures the unresolved mismatch.

### DIFF
--- a/src/attention_release_window.py
+++ b/src/attention_release_window.py
@@ -1,0 +1,6 @@
+from datetime import datetime
+
+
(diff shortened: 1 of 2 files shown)


fermano:
CI reproduces the boundary disagreement: `test_includes_event_at_end_boundary` fails because the implementation excludes the exact end timestamp. The linked customer samples require a decision on interval semantics, not just a test adjustment.

PR #360, opened 2026-07-13
Cache normalized escalation owners

fermano:
Caches normalized escalation-owner strings to reduce repeated routing work. The cache is process-local and explicitly clearable. A separate older draft approaches the same routing hot path with raw-key memoization.

### DIFF
--- a/src/attention_owner_cache.py
+++ b/src/attention_owner_cache.py
@@ -1,0 +1,12 @@
+_owner_cache: dict[str, str] = {}
+
+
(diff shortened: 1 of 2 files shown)


fermano:
Overlap review: https://github.com/fermano/TC1/pull/361 changes the same module but caches raw inputs and lacks invalidation. This PR has focused coverage and normalized keys; confirm whether the older draft contains any benchmark evidence worth preserving before treating it as a duplicate.

PR #361, opened 2026-07-13
Memoize incident owner resolution

fermano:
Draft restored from a May performance investigation. It memoizes by the raw owner string and has no invalidation hook or focused tests. The newer normalized-key proposal overlaps this code path but preserves different cache semantics.

### DIFF
--- a/src/attention_owner_cache.py
+++ b/src/attention_owner_cache.py
@@ -1,0 +1,7 @@
+_owner_cache: dict[str, str] = {}
+
+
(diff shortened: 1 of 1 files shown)


fermano:
This draft predates https://github.com/fermano/TC1/pull/360. The raw-key cache preserves the May benchmark shape, but it creates separate entries for equivalent owner spellings and has no clear hook. Keep it open only if the benchmark evidence cannot be represented in the newer proposal.

PR #362, opened 2026-07-13
Route blank owners to the default queue

fermano:
Normalizes blank and whitespace-only incident owners to the engineering-ops queue. No focused tests are included because the ownership contract is still being reconciled between intake and release operations.


fermano:
Blocking review note: please add focused cases for `None`, empty string, and whitespace-only owners, and document whether the default is owned by intake or release operations. Existing CI is green because this new path has no tests.

PR #363, opened 2026-07-13
Normalize release-channel aliases

fermano:
Normalizes stable and general-availability to ga. CI coverage is focused, but operational notes disagree on whether stable should map to ga or production for long-lived workers.


fermano:
Product/operations decision still open: new workers use `ga`, but rollback queries and the 4.6 worker use `production`. CI proves the proposed mapping is internally consistent; it does not establish the canonical audit label or migration owner.

PR #364, opened 2026-07-13
Remove legacy release-marker fallback

fermano:
Requires channel:version markers and removes the implicit stable-channel fallback. Repository tests pass, but mobile 4.6 rollout evidence still shows omitted-channel markers from one adapter.


fermano:
Compatibility blocker: mobile adapter 4.6 emitted 31 omitted-channel markers in the last rehearsal and has no recorded cutoff owner. Passing strict-parser tests are insufficient until the compatibility window or an adapter change is tracked.

PR #366, opened 2026-07-13
Stabilize handoff summary ordering

fermano:
Orders handoff rows by severity and owner for deterministic summaries. Named-owner coverage passes; operational samples also contain null owners, whose intended placement is not documented.


fermano:
Blocking edge case: production handoff rows may contain `owner: null`. The current key compares `None` with strings and the focused test covers named owners only. Define null-owner placement and add coverage before treating stable ordering as complete.

Issue #374, opened 2026-07-14
Filter delivery records by selected owners


fermano:
Add an optional backend filter for one or more owner names. Matching should use the same canonical owner semantics as routing and preserve input record order.

`owners=None` means no filtering. The behavior of an explicitly empty owner list is disputed between the API draft and dashboard representation and needs clarification.

PR #378, opened 2026-07-14
Prototype delivery owner filtering

fermano:
Prototype associated with https://github.com/fermano/TC1/issues/374. It normalizes selected owners and distinguishes None from an empty list, but it sorts matching records and therefore may not preserve source order.


fermano:
Prototype review: normalized matching is useful, but sorting changes delivery order. The current contract requires input order; the empty-list product decision and performance evidence were not resolved in this draft.

PR #380, opened 2026-07-14
EXP-132 Filter delivery records by canonical owner

fermano:
## What changed

- add an order-preserving backend owner filter
- reuse canonical routing keys for records and selections
- distinguish no filter from an explicit empty selection
- document and test deduplication, fallback matching, and ordering

Linear: https://linear.app/experttc2/issue/EXP-132/filter-delivery-records-by-canonical-owner

## Validation

- `python -m pytest -q`: 170 passed
- `git diff --check`: passed


fermano:
Changes requested: the implementation already canonicalizes blank selections to the fallback owner, but the public type and tests do not record that behavior. Please accept `None` in selected owner values and add focused coverage showing a blank selection matches blank-owner records through `engineering-ops`.

PR #382, opened 2026-07-15
Canonicalize equivalent delivery owner whitespace

fermano:
Collapses repeated and Unicode whitespace in delivery owner keys while preserving separators and the existing blank-owner default. Includes regression coverage for the duplicate-route reports.


fermano:
Post-implementation verification: CSV samples, Unicode whitespace, separators, and blank-owner fallback are covered. No remaining owner-normalization follow-up was identified.

fermano:

PR #384, opened 2026-07-15
Add opt-in source metadata to delivery summaries

fermano:
Adds backwards-compatible source metadata for delivery summaries. Default output remains unchanged; opted-in blank or missing values are omitted.


fermano:
Post-merge consumer review: the delivery-summary path is complete. The audit-export adapter still emits `source_id` and does not call this helper; deciding whether to map that adapter before the next audit export remains separate follow-up work.

fermano:

PR #385, opened 2026-07-15
Deduplicate delivery retries within a worker

fermano:
Adds a process-local delivery ID cursor that rejects repeated deliveries handled by the same worker. Focused tests cover retries and distinct IDs.


fermano:
Post-merge restart rehearsal: same-worker retries are fixed, but a restarted worker accepts the last delivered ID again because the cursor is process-local. Durable restart behavior remains unimplemented and needs explicit tracking.

fermano:

PR #386, opened 2026-07-15
Require canonical delivery queue names

fermano:
Rejects legacy queue aliases and accepts only canonical delivery queue names. Tests cover canonical normalization and the former general alias.


fermano:
Post-merge compatibility sample: mobile adapter 4.6 still sends `general` for roughly 8% of delivery handoffs. The earlier alias-removal Linear experiment was canceled; do not reopen it, but the remaining adapter migration needs suitable new or existing tracking.

fermano:

PR #387, opened 2026-07-15
Bound delivery batch submissions

fermano:
Rejects delivery batches above 100 records and covers the exact boundary. API callers are documented as responsible for chunking larger submissions.


fermano:
Post-merge call-site check: public API clients chunk batches, but the internal reconciliation scheduler still submits up to 250 records directly. Ownership is unclear between scheduler and delivery ingestion; follow-up is required before enabling the limit for that caller.

fermano:

PR #388, opened 2026-07-16
Preserve configured delivery lease grace

fermano:
Adds a small helper for selecting a configured delivery lease grace or the workspace default.

The branch currently follows the existing truthy-fallback convention. Review is requested on whether explicit zero has distinct operational meaning.

Validation: repository CI.

### DIFF
--- a/src/review_seed_lease_grace.py
+++ b/src/review_seed_lease_grace.py
@@ -1,0 +1,8 @@
+"""Lease-grace parsing proposed for the delivery worker."""
+
+DEFAULT_LEASE_GRACE_SECONDS = 30
+
+
+def effective_lease_grace(configured_seconds, default_seconds=DEFAULT_LEASE_GRACE_SECONDS):
+    """Return the configured grace period or the workspace default."""
+    return configured_seconds or default_seconds
--- a/tests/test_review_seed_lease_grace.py
+++ b/tests/test_review_seed_lease_grace.py
@@ -1,0 +1,18 @@
+import unittest
+
+from src.review_seed_lease_grace import (
+    DEFAULT_LEASE_GRACE_SECONDS,
+    effective_lease_grace,
+)
+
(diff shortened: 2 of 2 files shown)


fermano:
Blocking review finding: `configured_seconds or default_seconds` converts an explicit `0` into the workspace default. Zero disables the handoff grace period for drain-only queues, so this changes persisted policy meaning. Please distinguish `None` from `0` and add the explicitly requested zero-value regression case.

fermano:
Author follow-up: confirmed that zero is present in current workspace policy exports. The branch has not yet been updated; the implementation and focused regression requested above remain outstanding.

fermano (commented):
Review summary: explicit zero is a persisted policy value, not absence. The requested implementation change and focused zero-value regression remain outstanding.

fermano:
CI attribution note: this head reproduces the pre-existing ticket-workflow contract failure already tracked in Linear EXP-206. The observed failure is outside this PR's changed files, and no PR-specific failing test has been isolated from this run. Review readiness should consider the shared baseline separately from this PR's resolved or unresolved discussion.

fermano:
Coverage review: the new focused tests cover a missing value and a positive override, but not the explicitly supported `0` value identified in review. A regression test must assert that zero remains zero; the current suite would pass while the deployed policy behavior is still wrong.

fermano (commented):
Test-coverage review: explicit zero remains an uncovered supported value. The current default/positive tests are insufficient regression coverage.

fermano:
Related Linear record: [EXP-5](https://linear.app/experttc3/issue/EXP-5/validate-lease-grace-fallback-behavior) was written from the draft state. It records the lease-grace policy question, but does not yet include the latest partial test coverage or the review evidence confirming that explicit zero occurs in persisted policy exports.

PR #389, opened 2026-07-16
Honor Retry-After on throttled delivery handoffs

fermano:
Adds Retry-After parsing for delivery retries. Numeric delay values are supported and clamped to non-negative seconds.

A compatibility test also records the HTTP-date form observed from one upstream gateway; that case is not yet implemented.

Validation: repository CI, including the new Retry-After tests.

### DIFF
--- a/src/review_seed_retry_after.py
+++ b/src/review_seed_retry_after.py
@@ -1,0 +1,8 @@
+"""Retry-After parsing for delivery handoffs."""
+
+
+def parse_retry_after(value, now_epoch=None):
+    """Return a non-negative delay in seconds for a numeric Retry-After value."""
+    if value is None or not value.strip():
+        return None
+    return max(0, int(value.strip()))
--- a/tests/test_review_seed_retry_after.py
+++ b/tests/test_review_seed_retry_after.py
@@ -1,0 +1,18 @@
+import unittest
+
+from src.review_seed_retry_after import parse_retry_after
+
+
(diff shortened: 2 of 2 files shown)


fermano:
Review/CI finding: one gateway sends the RFC 7231 HTTP-date form of `Retry-After`, while `parse_retry_after` currently calls `int()` unconditionally. The new HTTP-date test records the observed header and fails. This is unresolved merge-blocking follow-up; numeric-only support is insufficient for the stated PR scope.

fermano (commented):
Review summary: the HTTP-date Retry-After path fails the submitted compatibility test and remains blocking for the stated gateway scope.

fermano:
CI attribution note: the full run stops during the pre-existing ticket-workflow collection failure tracked in Linear EXP-206, before the new Retry-After test executes. The HTTP-date parsing gap remains explicit in the submitted code and review discussion, but it should not be reported as a separately observed CI failure for this run.

fermano:
Coverage assessment: numeric and HTTP-date `Retry-After` inputs are both represented by automated tests. The HTTP-date case currently fails because the implementation calls `int()`; that is an implementation defect exposed by coverage, not missing or deferred test work. No test-coverage ticket is requested.

fermano (commented):
Test-coverage review: both accepted Retry-After forms have automated cases. The failing date case is implementation follow-up, not a coverage gap.

PR #395, opened 2026-07-17
Reject stale webhook timestamps

fermano:
Adds webhook timestamp freshness validation with coverage for recent, stale, future, and malformed values.

The future-timestamp case currently exposes an implementation defect in the proposed comparison.

Validation: repository CI, including four focused timestamp cases.

### DIFF
--- a/src/coverage_seed_webhook_time.py
+++ b/src/coverage_seed_webhook_time.py
@@ -1,0 +1,9 @@
+"""Freshness checks for signed webhook timestamps."""
+
+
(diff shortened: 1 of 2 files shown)


fermano:
Coverage assessment: recent, stale, future, and malformed timestamps all have focused automated cases. The future case currently fails and identifies a comparison bug, but adding another test would not improve path coverage. Treat this as implementation follow-up, outside this coverage audit.

fermano (commented):
Test-coverage review: all stated timestamp categories have automated cases. The red future case is an implementation defect, not missing coverage.

PR #397, opened 2026-07-17
Add opt-in tenant metadata to audit records

fermano:
Adds optional tenant metadata to projected audit records while intending to preserve the default record shape.

The initial focused test covers the opt-in path. Default compatibility coverage is under review.

Validation: repository CI.

### DIFF
--- a/src/coverage_seed_audit_projection.py
+++ b/src/coverage_seed_audit_projection.py
@@ -1,0 +1,8 @@
+"""Audit-record projection with opt-in tenant metadata."""
+
+
+def project_audit_record(event, include_tenant=False):
+    projected = {"id": event["id"], "action": event["action"]}
+    if include_tenant and event.get("tenant_id"):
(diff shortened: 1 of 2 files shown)


fermano:
Coverage review finding: the opt-in case is covered, but the compatibility-sensitive default path is not. Please add a regression asserting that `include_tenant=False` preserves the exact two-key record shape, plus a missing-tenant opt-in case, before treating coverage as complete.

fermano:
Coverage follow-up on the latest commit: the suite now covers the default two-key shape, opted-in tenant inclusion and trimming, and opted-in behavior when tenant metadata is absent. The earlier coverage finding is resolved; no Linear test follow-up is requested.

fermano (commented):
Test-coverage review on the latest commit: default compatibility, opted-in metadata, trimming, and absent metadata are covered. The earlier gap is resolved.

Issue #398, opened 2026-07-20
Audit export resume can omit events sharing a timestamp


fermano:
RC-55 reconciliation reproduced a gap when a page ended between two audit events with the same `occurred_at` value.

Observed sample:

- `evt-104` and `evt-105` both have `occurred_at=1784217600`.
- Page size was 1.
- The cursor returned after `evt-104` was `1784217600`.
- Resuming returned the next later timestamp and omitted `evt-105`.

The current cursor is persisted by scheduled reconciliation jobs. Those jobs already deduplicate received rows by event ID, but the repository contract does not explicitly state whether resume delivery is exactly-once or at-least-once.

Please preserve compatibility with persisted decimal cursors and add a regression that exercises a page boundary inside a same-timestamp group.

fermano:
Consumer contract check: the scheduled reconciliation importer uses event ID as an idempotency key and tolerates replayed rows. One repeated boundary row is acceptable; an omitted row is not. Cursor strings must remain decimal during the current mixed-version window.

fermano:
Linear tracking: [EXP-37](https://linear.app/experttc3/issue/EXP-37/inc-55-audit-export-resume-omits-same-timestamp-event) owns the implementation outcome. [EXP-46](https://linear.app/experttc3/issue/EXP-46/add-same-timestamp-audit-page-boundary-regression) records the focused regression requirement. Release coordination and raw-CSV presentation remain separate related records.

PR #399, opened 2026-07-20
Prototype offset cursors for audit exports

fermano:
Explores offset-based pagination to avoid dropping same-timestamp events in static exports. Includes a focused static-page test.

This does not yet establish behavior when records are inserted between pages or compatibility with persisted decimal timestamp cursors.

Related: #398

### DIFF
--- a/src/seed_audit_offset_cursor.py
+++ b/src/seed_audit_offset_cursor.py
@@ -1,0 +1,8 @@
+"""Offset-based audit export prototype."""
+
+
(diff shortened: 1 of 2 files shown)


fermano:
Review finding: offsets are position-dependent. If a retry inserts an earlier record between pages, the same offset can repeat one row and omit another. The prototype also cannot resume persisted decimal timestamp cursors without a migration layer. Keep this as exploratory; it is not suitable for the RC-55 recovery path.

PR #400, opened 2026-07-20
Prototype timestamp-and-ID audit cursors

fermano:
Explores a versioned timestamp-and-ID cursor that can resume inside a same-timestamp group and decode existing decimal cursors.

The branch does not make older workers understand newly emitted `v2:` cursors during rollback or mixed-version execution.

Related: #398

### DIFF
--- a/src/seed_audit_composite_cursor.py
+++ b/src/seed_audit_composite_cursor.py
@@ -1,0 +1,20 @@
+"""Composite audit cursor prototype."""
+
+
+def encode_composite_cursor(record):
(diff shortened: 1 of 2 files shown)


fermano:
Release review: new code can read the legacy decimal form, but rollback workers cannot read a `v2:` cursor written by a newer worker. Mixed-version reconciliation is still active for RC-55, so emitting the new format is not safe in this release. The ordering tests are useful; the cursor-format migration is superseded for the immediate recovery.

Issue #402, opened 2026-07-20
Replay reclaim can deliver billing webhook twice after worker restart


fermano:
## Observed in rc-56 validation

A billing replay for `stream=billing-eu`, `cursor=evt-8841` produced two external deliveries during a worker replacement.

Partial sequence reconstructed from service and gateway logs:

- worker-a recorded a claim and began dispatch
- worker-a was replaced before its completion update was visible
- worker-b accepted the same stream/cursor after comparing the stored lease with its local clock
- an old worker-a callback arrived after worker-b had taken ownership
- the gateway recorded two delivery IDs for the same invoice event

The two worker logs disagree by 89 seconds. The SQLite host timestamps and gateway receipt order agree that the second claim happened before the original 60-second lease had elapsed.

Current normal-path replay tests pass. The rc-56 promotion is held because the duplicate reached the billing sandbox. No production delivery or customer debit is confirmed.

## Scope evidence

The replay claim table persists across restarts. Both the CLI replay tool and background replay worker use `ReplayDispatcher`. The delivery gateway recorded no idempotency key for either attempt.

Useful prior work is being explored in separate draft branches. This issue records the incident rather than selecting one proposal.

PR #403, opened 2026-07-20
Draft: persist replay completion before delivery

fermano:
Explores closing the late-callback window by making the durable completion record precede the gateway call.

Observed locally:
- repeated dispatch after a completed row is skipped
- existing replay tests remain green
- no schema migration is required

This is based on #402. The branch is intentionally narrow; failure behavior between the database write and gateway acceptance still needs review.


fermano (commented):
The row-first ordering removes the duplicate in the included test, but a SIGKILL after the SQLite commit and before `gateway.send` would leave a delivered row without an external receipt. We do not currently have a reconciliation path for that state.

PR #404, opened 2026-07-20
Draft: serialize replay lease takeover

fermano:
Serializes claim inspection and takeover with `BEGIN IMMEDIATE` and verifies that an expired claim is recoverable through a new SQLite connection.

This preserves the current claim schema and caller-supplied timestamp interface. It addresses the overlapping-claim race seen while reproducing #402, although the incident logs include disagreement about the exact expiry time.

### DIFF
--- a/src/seed_replay_claims.py
+++ b/src/seed_replay_claims.py
@@ -22,42 +22,37 @@ class ReplayClaimStore:
         )
         self.connection.commit()
 
-    def claim(
-        self,
-        stream: str,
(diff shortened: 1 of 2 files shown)


fermano (commented):
The transaction closes the concurrent read/update window. In the rc-56 trace, however, worker-b supplied a timestamp 89 seconds ahead of the SQLite host. Please account for whether the lease comparison has a single authority across replacement workers.

PR #405, opened 2026-07-20
Draft: fence stale replay callbacks

fermano:
Adds a fence to replay claims and carries it through completion so an older callback cannot complete a newer claim.

The focused same-process takeover test passes. The migration is additive for the rolling 4.8 window described in the repository docs. Restart behavior and how fences should interact with gateway delivery identity remain open review questions for #402.

### DIFF
--- a/src/seed_replay_claims.py
+++ b/src/seed_replay_claims.py
@@ -1,4 +1,4 @@
-"""Durable claim records used by the replay dispatcher."""
+"""Durable replay claims with callback fencing."""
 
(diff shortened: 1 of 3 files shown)


fermano (commented):
The additive column is compatible with the 4.8 reader. The counter starts at zero in each `ReplayClaimStore`; please exercise a fresh process taking over a persisted row and an old callback arriving afterward. The gateway still sees no repeated delivery identity.

PR #441, opened 2026-07-25
Normalize copied owner labels before grouping

fermano:
Normalizes case and ordinary surrounding/internal whitespace before owner identities are grouped while leaving configured display labels unchanged.

Current regression coverage exercises ASCII case differences and normal spaces. Copied non-breaking-space and Unicode-equivalence examples were discussed but are not included in this branch.

The behavior is used by CSV and replay grouping paths.

### DIFF
--- a/src/seed_coverage_owner_unicode.py
+++ b/src/seed_coverage_owner_unicode.py
@@ -1,0 +1,6 @@
+"""Owner identity normalization used by delivery grouping."""
+
+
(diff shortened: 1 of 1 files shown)


fermano:
Coverage note: current examples exercise ASCII case and ordinary spaces. Please add a copied non-breaking-space or canonically equivalent Unicode owner label; this grouping path is also used by CSV import and replay.

PR #442, opened 2026-07-25
Add restart regression for retry identity

fermano:
Adds a deterministic fresh-connection regression for the supported SQLite-backed retry receipt path.

The cross-platform behavior is exercised by the active test. A Windows-specific handle-release fixture remains conditionally skipped because Windows is not in the repository CI matrix; it does not replace the active restart assertion.

### DIFF
--- a/tests/test_seed_retry_identity_restart.py
+++ b/tests/test_seed_retry_identity_restart.py
@@ -1,0 +1,26 @@
+import sqlite3
+import sys
+
+import pytest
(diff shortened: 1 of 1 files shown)


fermano:
The active fresh-connection test covers the supported restart behavior. The conditional Windows handle-release fixture is platform-specific; I do not see an additional coverage request unless Windows enters the supported CI matrix.

PR #443, opened 2026-07-25
Reject invalid batch sizes before enqueue

fermano:
Validates type, minimum, and configured maximum before a delivery batch is enqueued.

Focused tests cover zero, negative values, maximum-plus-one, both valid boundaries, booleans, and non-integer input. Review discussion still questions the chosen exception wording, but no uncovered behavior has been identified.


fermano:
The boundary matrix looks complete for this change. I would prefer a more specific exception message, but that is implementation feedback rather than a missing-test finding.

PR #444, opened 2026-07-25
Document preview payload metadata

fermano:
Documents the existing preview response fields used by support and the release runbook.

This branch changes documentation only. It does not alter runtime code, schemas, generated examples, or executable configuration.


fermano:
Confirmed this changes explanatory documentation only and mirrors the existing payload fields. No runtime or snapshot coverage is requested for this PR.

PR #445, opened 2026-07-25
Add optional owner filter to delivery summaries

fermano:
Adds an optional owner filter to delivery summaries.

Coverage status:
- no filter (None) is covered;
- populated filters, case folding, and ordering are covered;
- the empty-list contract is explicitly deferred and its regression test remains skipped pending a product decision on whether [] means “no owners” or “no restriction”.

That deferred case should remain tracked until the behavior is decided and the test can be enabled.

### DIFF
--- a/src/seed_owner_filter.py
+++ b/src/seed_owner_filter.py
@@ -1,0 +1,14 @@
+"""Owner filtering used by delivery-summary previews."""
+
+
+def filter_delivery_records(records, owners=None):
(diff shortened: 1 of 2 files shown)


fermano:
The skipped empty-list case is intentionally deferred, but it leaves a real contract unprotected. Please track a decision on whether [] means no owners or no restriction, then enable the regression test.

PR #446, opened 2026-07-26
Carry owner context into resumed delivery

fermano:
Carries the operator-selected owner value from a queued delivery into a resumed attempt.

This path is used after timeout recovery and manual replay. The queue representation uses `None` when no override was supplied; values may originate in CSV imports or operator edits.

### DIFF
--- a/src/retry_owner_context.py
+++ b/src/retry_owner_context.py
@@ -1,0 +1,6 @@
+"""Owner selection for delayed delivery attempts."""
+
+
+def owner_for_attempt(original_owner, owner_override=None):
+    """Return the owner recorded for a resumed attempt."""
+    return owner_override or original_owner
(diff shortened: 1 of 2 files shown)


src/retry_owner_context.py
```diff
@@ -0,0 +1,6 @@
+"""Owner selection for delayed delivery attempts."""
+
+
+def owner_for_attempt(original_owner, owner_override=None):
+    """Return the owner recorded for a resumed attempt."""
+    return owner_override or original_owner
```
fermano, line 6: An explicit empty owner is used to clear assignment during operator replay. `or` treats it as absent and restores the old owner. Please distinguish `None` from an intentional empty value before this is merged.

fermano (commented):
One replay edge case still changes the recorded owner.

fermano:
I verified that CSV import currently trims blank owner cells, but manual replay can still submit an explicit empty value. I have not changed that behavior on this branch; the review question remains open.

PR #447, opened 2026-07-26
Accept cursor aliases during client rollout

fermano:
Accepts both pagination field names while dashboard and worker clients roll independently.

The canonical request field is `cursor`; `after` remains accepted for clients that have not yet been upgraded.

### DIFF
--- a/src/cursor_alias.py
+++ b/src/cursor_alias.py
@@ -3,4 +3,6 @@
 
 def select_cursor(parameters):
     """Read a cursor while old and new clients overlap."""
-    return parameters.get("after") or parameters.get("cursor")
+    if "cursor" in parameters:
+        return parameters["cursor"]
+    return parameters.get("after")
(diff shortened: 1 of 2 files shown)


src/cursor_alias.py
```diff
@@ -0,0 +1,6 @@
+"""Compatibility handling for paginated delivery cursors."""
+
+
+def select_cursor(parameters):
+    """Read a cursor while old and new clients overlap."""
+    return parameters.get("after") or parameters.get("cursor")
```
fermano: When both fields arrive during a rolling client upgrade, `cursor` is canonical and must win. Preferring `after` can replay an older page.

fermano (commented):
The overlap case needs a deterministic winner.

fermano:
Updated the current head so the canonical field wins when both names are present, and added the overlap regression. Rechecked the original concern against the new test; no review follow-up remains from that thread.

PR #448, opened 2026-07-26
Show trace source beside delivery previews

fermano:
Adds the compact trace label shown beside delivery preview rows.

The value is presentation-only and is derived from the existing source and trace identifier fields; no persisted payload shape changes.


fermano (commented):
The source fallback and presentation-only scope look consistent. The colon separator is a style preference, not a merge blocker; no change requested.

PR #449, opened 2026-07-26
Surface failure context in grouped dispatch

fermano:
Adds an error-context value to grouped delivery summaries so operators do not have to open each attempt separately.

The summary remains empty for successful batches and returns the available failure text when a delivery attempt fails.

### DIFF
--- a/src/batch_failure_context.py
+++ b/src/batch_failure_context.py
@@ -1,0 +1,10 @@
+"""Summaries for grouped delivery attempts."""
+
+
+def batch_failure_context(results):
+    """Return the error context displayed for a failed batch."""
+    failure = None
+    for result in results:
+        if not result["ok"]:
+            failure = result.get("error")
+    return failure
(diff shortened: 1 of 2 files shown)


fermano (commented):
The summary currently loses the failure that operators need for correlation.

src/batch_failure_context.py
```diff
@@ -0,0 +1,10 @@
+"""Summaries for grouped delivery attempts."""
+
+
+def batch_failure_context(results):
+    """Return the error context displayed for a failed batch."""
+    failure = None
+    for result in results:
+        if not result["ok"]:
+            failure = result.get("error")
```
fermano, line 9: This overwrites the first failure with every later one. Incident and retry correlation use the first failed attempt as the stable reference; preserve it and verify a batch containing two different failures.

fermano:
The existing summary happened to display the last failure, while the incident timeline and retry lookup use the first failed attempt. No follow-up commit has been made yet because the stable operator contract needs to win over the incidental display behavior.

PR #450, opened 2026-07-26
Apply release target to queued promotions

fermano:
Uses release target information when applying promotion actions that may have waited in the control-plane queue.

Actions produced by older clients can omit the target. Newer actions include the target that was visible when the request was prepared.

### DIFF
--- a/src/promotion_target.py
+++ b/src/promotion_target.py
@@ -1,0 +1,6 @@
+"""Target selection for queued promotion actions."""
+
+
+def target_for_promotion(action, current_target):
+    """Choose the deployment target when a queued action is applied."""
+    return action.get("target") or current_target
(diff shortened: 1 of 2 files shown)


src/promotion_target.py
```diff
@@ -0,0 +1,6 @@
+"""Target selection for queued promotion actions."""
+
+
+def target_for_promotion(action, current_target):
+    """Choose the deployment target when a queued action is applied."""
+    return action.get("target") or current_target
```
fermano, line 6: A queued action can carry the target from before a release switch. Reusing it without comparing the action generation to the current release generation can promote the old slot. Please re-read or reject stale actions rather than trusting the stored target.

fermano (commented):
Queued promotions still need a fence against target changes.

fermano:
I initially treated the target stored on the action as authoritative for rollback, but release handoff can occur while the action waits. The generation/fencing point in review is not addressed on the current head.

Issue #487, opened 2026-08-12
Manifest checksum changed after export row count changed


fermano:
The RC-66 package manifest checksum changed between two local builds while people were testing adjustment rows.

Post-PR #488 status: package contents were fixed first. PR #488 added the approved `partner_credit` row boundary and merged into `release/rc-66` as `dc9f951bde74a7a29f51b7b50129ad7161e8e3ef`.

Keep this open only for fresh evidence that manifest generation is still unstable after the expected package row-count change. A checksum change caused by adding the approved row is expected content drift, not a standalone hashing defect.

fermano:
Checksum changed after row count changed, so it may be an effect. I would check package contents before treating manifest hashing as its own bug.

fermano:
PR #488 fixed the known package-content mismatch before treating checksum behavior as a standalone defect. Since row-count changes can legitimately move the manifest checksum, this should stay open only for fresh evidence that the checksum remains unstable after the `partner_credit` package row boundary is applied.

Issue #486, opened 2026-08-12
Nova asked whether all adjustment reasons can be exported


fermano:
Nova pilot feedback asked whether their daily CSV can include every adjustment reason, including internal corrections and rebills.

Post-PR #488 status: this remains a product/contract question. PR #488 deliberately did not export all adjustment reasons; the merged RC-66 package selector allows only package-approved `partner_credit` adjustment rows.

Do not silently broaden this while maintaining the Atlas package fix.

fermano:
This is probably not something to sneak into the RC-66 fix. Nova's request sounds contract-shaped, not just implementation cleanup.

fermano:
PR #488 deliberately leaves this policy question open. The merged RC-66 package fix allows only package-approved `partner_credit` adjustment rows; it does not export all adjustment reasons, `internal_correction`, or `rebill` for Nova.

Issue #484, opened 2026-08-12
Preview CSV can show adjustment twice after dry-run reset


fermano:
A preview CSV generated during a dry-run reset showed two adjustment-looking lines for Atlas.

Post-PR #488 status: this is not the resolved Atlas final-package root cause. PR #488 fixed the uploaded package selector for posted `partner_credit` adjustment rows and intentionally did not change preview reset behavior.

Keep this open only for preview-specific evidence that duplicate rows survive in preview output after dry-run reset.

fermano:
I could only reproduce the double row in preview after `dry_run=true`. The uploaded partner package did not have this duplicate, but it happened around the same time as the short Atlas file.

fermano:
PR #488 intentionally did not close this preview-specific note. The merged package selector fix covers the final uploaded CSV path for `partner_credit` adjustment rows, while this issue still needs separate evidence if the preview duplicate after `dry_run=true` / reset survives into preview behavior. It should not be treated as the Atlas package root cause.

Issue #483, opened 2026-08-12
Atlas pilot package short one reconciliation row


fermano:
Atlas pilot export package from the overnight RC-66 run did not match the finance reconciliation sheet.

Observed in the final partner CSV package:

- `ship-918` exported
- `retry-223` exported
- `adj-772` absent

The internal row for the missing line looked like:

```json
{
  "tenant_id": "atlas-pilot",
  "partner": "atlas",
  "external_id": "adj-772",
  "event_type": "adjustment",
  "state": "posted",
  "reason": "partner_credit",
  "amount_cents": -640
}
```

There was also a local draft CSV where an `internal_correction` row appeared, but that draft was not the partner package Atlas saw. I do not know whether this belongs in export selection or in the finance mapper yet.

fermano:
Finance follow-up from the pilot call: the `partner_credit` line was expected in the Atlas package. The local draft that showed `internal_correction` was from a different spreadsheet export; I don't think Atlas should receive that one, but the naming is messy.

fermano:
Resolved by PR #488, merged into `release/rc-66` as `dc9f951bde74a7a29f51b7b50129ad7161e8e3ef`.

The final package selector now includes posted `adjustment` rows only when `reason=partner_credit`, covering the missing Atlas `adj-772` row without exporting `internal_correction` or broader adjustment reasons. Validation: GitHub Actions CI run #529 passed on PR head `11e098e27e5b3338991e9b97cca714ecb8d3ef83`; local focused RC-66 tests passed (5); local full suite passed (210).

PR #488, opened 2026-08-12
Gate RC-66 adjustment export reasons

fermano:
Narrows the RC-66 partner export pilot so it fixes the Atlas package mismatch without deciding Nova's broader all-adjustment request.

What changed:
- posted `adjustment` rows now export only when the persisted reason is `partner_credit`;
- `internal_correction`, `rebill`, and adjustments without an approved package reason stay out of the final partner CSV;
- RC-66 partner export notes now document that boundary.

Why:
- GitHub #483 / EXP-191 show the uploaded Atlas package was missing `adj-772` with `reason=partner_credit`;
- the preview duplicate screenshot (#484 / EXP-192), Nova all-reasons question (#486 / EXP-194), and checksum drift (#487 / EXP-195) remain separate from the package selector fix.

Validation:
- `/private/tmp/TC1-rc65-py312-venv/bin/python -m pytest tests/test_rc66_partner_export.py` -> 5 passed
- `/private/tmp/TC1-rc65-py312-venv/bin/python -m pytest` -> 210 passed

### DIFF
--- a/docs/rc66-partner-export.md
+++ b/docs/rc66-partner-export.md
@@ -7,5 +7,10 @@ Support and finance notes sometimes use "credit", "correction", "rebill", and
 "adjustment" loosely. Treat persisted row fields as authoritative when deciding
 whether a row belongs in the export package.
 
+For RC-66, posted `adjustment` rows are package-eligible only when the persisted
+reason is `partner_credit`. Internal corrections, rebills, and adjustments
+without a package-approved reason stay out of the final partner CSV until the
+contract language changes.
+
 Dry-run rows are visible in preview output but should not appear in the final
 partner CSV.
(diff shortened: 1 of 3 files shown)


fermano:
This branch gets the Atlas credit row into the candidate list, but please look at the reason values before merging. The pilot thread has `partner_credit` and `internal_correction` mixed together, and I don't think they mean the same thing for the partner CSV.

fermano:
Updated in `11e098e`: the selector now exports posted `adjustment` rows only for `reason=partner_credit`, and explicitly keeps `internal_correction`, `rebill`, and missing-reason adjustments out of the final partner CSV. Local validation passed: focused RC-66 partner export tests (5) and the full suite (210) under Python 3.12.

This resolves the reason-value concern for the focused Atlas package scope, while leaving Nova's all-reasons policy question and the preview/checksum notes separate.

PR #490, opened 2026-08-12
Backport visible receipt windows to RC-67

fermano:
Backports the RC-67 receipt-window selector fix from `main` onto `release/rc-67` and keeps the release note aligned with the artifact readout.

What changed:
- package selection now uses `visible_at` when a receipt replay supplies it, falling back to `created_at` for legacy rows;
- package rows now carry both `created_at` and `visible_at` so the readout can show which timestamp controlled inclusion;
- added regressions for a late partner-visible receipt entering the 100-120 window, a replay-visible receipt moving out of that window, and legacy created-at behavior;
- clarified that `rc67-atlas-receipts-20260812-b` can remain stale and must be rebuilt/read out from the release branch before the package is called clean.

Validation:
- Known Atlas sample probe on this branch returned both `base-114` and `late-419` for the 100-120 window.
- `/private/tmp/TC1-rc65-py312-venv/bin/python -m pytest tests/test_rc67_batch_window.py -q` -> 4 passed.
- `/private/tmp/TC1-rc65-py312-venv/bin/python -m pytest -q` -> 209 passed, 3 subtests passed.

Release posture:
- This addresses the branch/selector side of #492 / EXP-197 / EXP-198.
- Do not mark RC-67 unblocked until this PR is merged into `release/rc-67` and a rebuilt RC artifact verifies both `base-114` and `late-419`.
- #493 should remain a checksum follow-up only if the manifest is still unstable after the row set is settled.
- #494 is documentation cleanup following package behavior; #495 remains a separate policy question.

### DIFF
--- a/docs/rc67-batch-window.md
+++ b/docs/rc67-batch-window.md
@@ -6,3 +6,9 @@ window is normally expressed using the receipt's visible `created_at` timestamp.
 Late acknowledgements and partner replays may attach additional timestamps. When
 those values are present, release-room checks compare package contents against
 the timestamp that best represents when the row became visible to the partner.
+
+## Release-room note
+
+The main branch has a receipt-window change under review/landed depending on
+which dashboard is checked. The release candidate package still needs a readout
+against the branch that actually builds the RC-67 artifact.
--- a/docs/rc67-batch-window.md
+++ b/docs/rc67-batch-window.md
@@ -1,7 +1,8 @@
 # RC-67 batch window
 
 Receipt packages include final receipt rows for a half-open package window. The
-window is normally expressed using the receipt's visible `created_at` timestamp.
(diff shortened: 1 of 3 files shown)


fermano:
Update after branch reconciliation: this PR is no longer note-only. It now carries the `visible_at` selector fix and regressions, passed PR Actions CI run #538 on head `47af12b47f9e03913b9e401819b5b7e1a6832818`, and was merged into `release/rc-67` as `95ddd7f1b0a57bb3cd8afa92c91bedbdb1c6519a`.

Release posture remains held until a rebuilt RC-67 artifact from `release/rc-67` verifies both `base-114` and `late-419`; the old `rc67-atlas-receipts-20260812-b` readout should be treated as stale, not proof that the merged branch is still wrong.

Issue #508, opened 2026-08-12
RC-69 release branch may need settlement status alias backport


fermano:
The release branch was cut before the mainline `settled-final` handling landed in #507.

Early release-room notes also mention a staging-only `preview-settled` row from a dry-run screenshot, but it is unclear whether that row belongs in the final settlement package.

Relevant history:

- #506 / #507 fixed `settled-final` on `main`.
- The final artifact identity check is tracked separately; a matching row count does not prove the artifact came from the current release branch.

Do not use this issue alone to declare RC-69 ready. If a release-branch PR continues, it needs to avoid broadening final-package eligibility from preview-only evidence.

PR #509, opened 2026-08-12
Backport settlement status aliases for RC-69

fermano:
Refs #508.

Backports status aliases seen during RC-69 settlement dry runs.

Included states:

- `settled-final`, which was fixed on main in #507.
- `preview-settled`, which appeared in one staging screenshot during the same release-room window.

This has not been reconciled with the final artifact identity check or with the later manifest source-ref notes. Treat as a candidate branch, not proof that RC-69 is ready.

### DIFF
--- a/src/rc69_settlement_window.py
+++ b/src/rc69_settlement_window.py
@@ -1,6 +1,6 @@
 """Settlement-row selection used by the RC-69 package readout."""
 
-READY_STATES = {"posted", "settled"}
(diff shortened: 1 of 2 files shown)


fermano:
Release-room correction after this PR was opened: the newer west readout has the expected row count, so the current hold is artifact identity/provenance (#510), not missing rows. This branch also includes `preview-settled`, which came from preview evidence and should not be accepted into final package eligibility without a separate decision.

PR #519, opened 2026-08-12
Prototype RC-71 route-scoped invoice keys

fermano:
## Summary

Prototype from the first RC-71 route-ledger investigation. It changes invoice collapse from tenant+invoice to tenant+route+invoice so the same invoice can appear on multiple partner routes.

## Caveat

This branch was written against a QA sample that used `partner_route`. Later release-room notes may use `route_id` from the real adapter. It also does not address later `voided`/`retracted` events; it still keeps the first exportable row it sees.

Useful part: route-scoped identity.
Not proven: final RC-71 route field name, per-route retraction behavior, or candidate artifact row count.

### DIFF
--- a/src/rc71_route_manifest.py
+++ b/src/rc71_route_manifest.py
@@ -9,7 +9,7 @@ def _normalize(value):
 
 
 def _route_id(event):
-    return event.get("route_id") or event.get("partner_id") or "primary"
+    return event.get("partner_route") or event.get("partner_id") or "primary"
 
 
 def _sequence(event):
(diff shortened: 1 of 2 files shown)


fermano:
Route split looks like the right shape, but this was tested with `partner_route` from QA. Before using it for RC-71, reconcile with the final artifact field (`route_id`) and the later voided card event.

fermano:
Superseded by PR #526, merged into `release/rc-71` as `5b5cfdde1945d208baf3c0166ca2a0d99bc6ba79`. The useful route-scoped identity idea was carried forward, but reconciled to current `route_id` / legacy `partner_id` and combined with route-local retraction handling. Closing this prototype unmerged so it is not mistaken for an alternate unblock.

PR #520, opened 2026-08-12
Prototype RC-71 sequence-based invoice retractions

fermano:
## Summary

Prototype from the second RC-71 investigation. It chooses the highest sequence/event_sequence for an invoice and drops the row when the latest event is `voided`, `retracted`, or canceled.

## Caveat

This branch keys by tenant+invoice only. If the same invoice is present on two partner routes, a retraction on one route can suppress the other route. It should not be merged as-is unless the final artifact proves route-local identity is irrelevant.

Useful part: sequence-based retraction semantics.
Not proven: route-local isolation.

### DIFF
--- a/src/rc71_route_manifest.py
+++ b/src/rc71_route_manifest.py
@@ -20,12 +20,7 @@ def _sequence(event):
 
 
 def manifest_rows(events):
-    """Return exportable invoice rows in first-seen order.
-
-    The release manifest is intentionally conservative: only ready/posted rows are
-    emitted, unsupported states are ignored, and duplicate invoice rows are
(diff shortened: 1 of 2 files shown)


fermano:
The single-route void case matches the first reproduction. This still needs a route-local check before it can be used for RC-71 because the same invoice appears on card and bank routes in #522.

fermano:
Superseded by PR #526, merged into `release/rc-71` as `5b5cfdde1945d208baf3c0166ca2a0d99bc6ba79`. The useful sequence-based retraction behavior was carried forward, but scoped to route+invoice so a card-route void cannot suppress the bank route. Closing this prototype unmerged.

PR #521, opened 2026-08-12
Add RC-71 partner_route alias notes

fermano:
## Summary

Small cleanup from a QA export that called the route field `partner_route`.

## Caveat

This does not address the active package behavior if the release artifact uses `route_id`; it also keeps first-seen invoice collapse and does not handle later retractions. Treat as context/history, not the release unblock by itself.

Useful part: records the QA alias.
Not proven: active release adapter shape or route-local void behavior.

### DIFF
--- a/docs/rc71-route-manifest.md
+++ b/docs/rc71-route-manifest.md
@@ -2,12 +2,13 @@
 
 RC-71 route invoice packaging emits invoice rows for partner routes from the release branch artifact.
 
-Current field names expected from the release artifact:
+Current field names seen during the release-room window:
 
 - `tenant_id`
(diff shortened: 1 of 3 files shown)


fermano:
Useful docs cleanup for the QA alias, but release-room evidence has not confirmed `partner_route` in the actual RC-71 artifact. This should not be merged as the unblock by itself.

fermano:
Closing as stale QA-alias context. PR #526 resolved the active RC-71 route/retraction behavior without promoting the unconfirmed `partner_route` field into the release path. Keep #523/EXP-229 as context only unless a future final artifact actually emits `partner_route`.

PR #527, opened 2026-08-12
Prototype recursive admin export redaction

fermano:
This carries the sanitizer through nested connector/auth structures so support/admin export summaries do not expose connector credentials.

Notes from the prototype pass:
- Redacts nested `access_token` and `client_secret` under connector auth blocks.
- Preserves the existing non-mutating copy behavior for the examples covered here.
- Still uses broad fragment matching, so fields like `secretary_email` and `public_token_hint` need a closer read before this is safe to merge.

I’m leaving this open because it may be useful history, not necessarily the final shape.

### DIFF
--- a/src/tc1_export_sanitizer.py
+++ b/src/tc1_export_sanitizer.py
@@ -13,21 +13,20 @@ def _is_sensitive_key(key):
     return any(fragment in normalized for fragment in SENSITIVE_FRAGMENTS)
 
 
+def _sanitize_value(value):
+    if isinstance(value, dict):
+        sanitized = {}
+        for key, child in value.items():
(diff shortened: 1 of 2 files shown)


fermano:
Checked against the Atlas sample from the admin bundle thread. This branch does stop nested auth from showing up, but it also hides `secretary_email` and `public_token_hint`. I would not merge it as-is unless the safe-field behavior is reconciled.

fermano:
Closing as superseded by PR #534, merged to `main` as `ba0cb0f`. #534 keeps the useful recursive coverage from this prototype while replacing broad fragment matching with exact secret-key matching so safe labels remain visible.

PR #528, opened 2026-08-12
Tighten admin export secret key matching

fermano:
This narrows the admin export sanitizer from broad fragment matching to exact normalized secret keys.

Why this exists:
- Avoids hiding safe fields such as `secretary_email` and `public_token_hint`.
- Keeps the current non-mutating sanitizer behavior.
- Does not yet prove nested connector auth is redacted, so this may only be a partial fix for the current report.

Leaving as open implementation history until the active TC1 work is reconciled.

### DIFF
--- a/src/tc1_export_sanitizer.py
+++ b/src/tc1_export_sanitizer.py
@@ -1,7 +1,15 @@
 """Helpers for preparing TC1 admin export bundles."""
 
 REDACTION = "[redacted]"
-SENSITIVE_FRAGMENTS = {"password", "token", "secret", "api_key", "private_key"}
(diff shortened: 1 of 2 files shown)


fermano:
This is closer for the safe-label part and keeps the source payload intact. I still do not see coverage for `connectors[].auth.access_token`, `client_secret`, or `refresh_token`, which is the piece that reopened the admin-bundle report.

fermano:
Closing as superseded by PR #534, merged to `main` as `ba0cb0f`. #534 keeps the useful exact-key/safe-label behavior from this prototype and extends it recursively to nested connector auth.

PR #529, opened 2026-08-12
Sanitize admin export preview payloads recursively

fermano:
This makes admin export preview sanitization recursive and exact-key based.

What it covers:
- Nested connector auth keys like `access_token` and `client_secret`.
- Safe labels like `secretary_email` and `public_token_hint`.

Concern:
- The implementation mutates the parsed payload in place. That may be fine for a one-off preview render, but the audit preview and support summary can share the same object in the current request path.

Leaving open until the mutation concern is reconciled with the current report.

### DIFF
--- a/src/tc1_export_sanitizer.py
+++ b/src/tc1_export_sanitizer.py
@@ -1,7 +1,15 @@
 """Helpers for preparing TC1 admin export bundles."""
 
 REDACTION = "[redacted]"
-SENSITIVE_FRAGMENTS = {"password", "token", "secret", "api_key", "private_key"}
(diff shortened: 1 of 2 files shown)


fermano:
The recursive exact-key behavior looks like the right direction, but mutating the parsed bundle is not safe for the audit preview path. The final version should probably combine this key policy with the copy behavior from the existing sanitizer.

fermano:
Closing as superseded by PR #534, merged to `main` as `ba0cb0f`. #534 keeps nested exact-key redaction while avoiding in-place mutation of the shared parsed payload.

PR #540, opened 2026-08-13
Exclude disabled workspaces from retry digest queue

fermano:
Prevents disabled or deleted workspaces from entering the retry digest queue.

Known gap: the branch has no test for the disabled-but-retry-enabled workspace from the release-room note. The author mentioned local validation in Slack, but no repository evidence is attached here.

### DIFF
--- a/src/tc1_digest_disabled_workspace.py
+++ b/src/tc1_digest_disabled_workspace.py
@@ -1,0 +1,9 @@
+"""Retry digest filters for disabled TC1 workspaces."""
+
+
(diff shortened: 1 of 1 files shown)


fermano:
Triage note: CI/coverage evidence is incomplete here. The branch handles `disabled` and `deleted_at`, but there is still no repository test for `disabled=True` with `retry_enabled=True`, which is the release-room shape that put disabled workspaces back into the digest queue last week.

PR #541, opened 2026-08-13
Document drain replay ownership handoff

fermano:
Docs-only clarification for drain replay ownership.

This is context for operators and does not touch runtime paths. It should not create product-defect tracking unless another signal links it to a current blocker.


fermano:
Triage note: this is docs-only context. It supersedes some older RC notes about ownership handoff, but I do not see a product defect here unless another current signal links the wording to runtime behavior.

PR #542, opened 2026-08-13
Preserve checkpoint rows during digest cleanup

fermano:
Retry digest cleanup currently deletes retained checkpoint rows too aggressively in the replay path.

This branch adds a small retention guard. It does not yet include a regression test for the replay-retained row that failed in the operator notes.

Related context is scattered: an older Linear item says this was done, but that note appears to predate retained replay rows.


fermano:
Blocking-ish note: the retained replay sample `atlas/retry-774 retained_for_replay=True state=pending age=96` is still not covered by a regression test. There is an older Linear item saying digest cleanup is done, but that evidence appears to predate retained replay rows.

PR #544, opened 2026-08-13
Quarantine py312 digest ordering check

fermano:
Adds a skip around the Python 3.12 digest ordering check.

This is intentionally a quarantine branch, not a root-cause fix. The question for tracking is whether there is already a Linear item for the underlying ordering instability.


fermano:
Review note: please don't merge the skip as the final fix. The underlying Python 3.12 digest ordering instability still needs owner/tracking if no current Linear item already covers it.

PR #543, opened 2026-08-13
Normalize hold reason labels for operator view

fermano:
Normalizes operator-facing hold reason labels so `customer-paused` and `customer_paused` render consistently.

Potential overlap: PRs and issues around digest/escalation reason values may be related, but this branch is focused on operator display labels.

PR #545, opened 2026-08-13
Normalize escalation reason before digest fanout

fermano:
Normalizes escalation reason values before retry digest fanout.

This may overlap with the operator label PR, but the fanout value is machine-facing while the operator label is display-facing. Needs triage to avoid duplicate work or accidental divergence.


fermano:
Triage note: potential duplicate with #543, but not identical. This branch changes the machine-facing digest reason value, while #543 changes the operator label. One tracker should call out the split to avoid divergent normalizers.

PR #546, opened 2026-08-13
Backfill legacy token on export replay

fermano:
Uses `legacy_token` when replay records do not yet contain `export_token`.

The branch has no tests for the EU dry-run sample mentioned in the issue thread. There is an old canceled Linear issue with a similar title; it may or may not be the correct tracker.

### DIFF
--- a/src/tc1_legacy_export_token.py
+++ b/src/tc1_legacy_export_token.py
@@ -1,0 +1,9 @@
+"""Legacy export replay token normalization."""
+
+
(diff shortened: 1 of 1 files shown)


fermano:
Triage note: the EU dry-run sample used `legacy_token` without `export_token`; this branch handles that shape but has no regression coverage for it. The old Linear cleanup item was canceled before EU evidence arrived, so match carefully.

PR #547, opened 2026-08-13
Show ownerless retry batches in dashboard

fermano:
Shows ownerless retry batches in the operator dashboard, using region alias as a fallback label.

This is related to an open support report about ownerless digest cards, but the product owner has not confirmed whether `region_alias` is the desired final display value.


fermano:
Triage note: the `ready-to-merge` label came from a dashboard smoke check, but support has not confirmed whether `region_alias` is an acceptable owner fallback. This may be tracked already under a differently named ownerless-card item.

PR #555, opened 2026-08-13
Use receipt probe for retry preview count

fermano:
Switches the retry preview count to use receipt rows when they are available.

Context from the release desk was inconsistent: one note said the daily digest total was correct while the preview card was low; another said the preview was only showing delivered receipts and omitting pending receipts. This branch adds a helper, but the calling path is not wired in this diff.

No reviewer is assigned yet.


fermano:
The preview count looked low on the 09:40 readout, but the digest total matched by 09:52. I could not tell whether that was delayed receipts, pending rows being hidden, or just the preview path not calling this helper yet.

PR #556, opened 2026-08-13
Add retry card glossary examples

fermano:
Adds operator glossary examples for retry cards, owner labels, region aliases, and receipt probe counts.

The glossary includes examples copied from recent screenshots, but does not change runtime behavior.


fermano:
Glossary copy uses `region alias` and `receipt probe` examples from support screenshots. It may explain the terms, but it does not settle whether the screenshots describe current behavior or stale examples.

PR #557, opened 2026-08-13
Simplify disabled workspace replay selection

fermano:
Simplifies replay candidate selection for digest backfill.

The branch keeps deleted workspaces out, but does not check the `disabled` flag. The author note says disabled accounts should already be filtered before replay selection. The daily digest queue and replay backfill may not use the same entry point.


fermano:
This removes the disabled guard from replay selection, not from the daily queue selector. I don't know whether the two selectors can diverge safely; the disabled workspace incident mentioned both queue and replay in different places.

PR #558, opened 2026-08-13
Handle partner paused hold reason

fermano:
Adds alias handling for `partner-paused` and `partner_paused` in a partner hold helper.

This is similar to the older customer-paused work, but the partner case came from a later digest CSV sample. Operator labels were not mentioned in that sample.


fermano:
The partner-paused sample came from a CSV export, not the operator card. It shares vocabulary with customer-paused, but I don't know whether the owner should be the same workstream.

PR #559, opened 2026-08-13
Require explicit export token on replay rows

fermano:
Moves replay rows toward explicit `export_token` usage and returns an empty token when it is absent.

The motivation was to stop treating old staging fixtures as canonical. The EU dry-run sample from earlier this week is not mentioned in the code comment or tests.


fermano:
If the EU replay row with only `legacy_token` is still valid, this would emit an empty token. If that sample was a staging artifact, the stricter behavior may be fine. I don't see that distinction in this PR.

PR #560, opened 2026-08-13
Pin digest ordering plugin behavior

fermano:
Adds a small ordering test around the digest plugin behavior seen on Python 3.12.

The PR does not remove the existing quarantine. The author reported a clean local run, but no linked CI check is mentioned in the description.


fermano:
This test is green locally according to the author, but the quarantine marker from the earlier py312 PR is still present elsewhere. Passing this small test may not mean the original ordering flake is gone.

PR #571, opened 2026-08-13
Preserve cents in Meridian export preview

fermano:
Adds a small helper for preview amount/tax cents in the Meridian export path.

Related release note: RC-73 readout showed `preview_vat=19` while the CSV retained `19.37`. This branch covers cents helpers but does not show where the preview renderer calls them yet.

No Linear link was included in the release note.


fermano:
Release readout note says the CSV total was correct and only the preview card rounded VAT. This PR adds helpers but I don't see the preview renderer changed in the diff.

PR #572, opened 2026-08-13
Closed: superseded by route-scoped RC74 zero dispatch fix

fermano:
Prototype from the Lagoon incident.

It keeps explicit `delay_seconds=0` / `send_after_seconds=0` as a real immediate-dispatch override instead of falling back to the workspace default.

This was written before RC-74 moved dispatch records to route-scoped `scope_key` output, so the branch still returns a tenant-level shape. The Lagoon note did not mention route IDs when this was opened.

### DIFF
--- a/src/tc1_lagoon_dispatch_delay.py
+++ b/src/tc1_lagoon_dispatch_delay.py
@@ -1,0 +1,21 @@
+"""Lagoon dispatch delay override prototype.
+
+This prototype came from the original Lagoon customer incident. It preserves an
+explicit zero-delay override, but it uses the old tenant-level output shape.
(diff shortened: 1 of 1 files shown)


fermano:
Superseded by PR #579, merged into `release/rc-74` as `c4495a596d52b119288271b6c2cc1d3b059931b8`.

The useful behavior from this prototype was explicit-zero preservation for `send_after_seconds=0` / `delay_seconds=0`. #579 keeps that behavior, but applies it to the current RC-74 route-scoped `scope_key` / `window_seconds` artifact shape instead of this branch's stale tenant-level output.

PR #573, opened 2026-08-13
Closed: route-scope tests included in RC74 zero dispatch fix

fermano:
Adds tests around the RC-74 route-scoped dispatch window helper.

This is based on the current release branch shape: `scope_key`, route identity order, and `window_seconds`. It covers route scoping and blank fallback, but it does not cover Lagoon's explicit zero-delay override from the earlier incident thread.

### DIFF
--- a/tests/test_tc1_lagoon_release_window.py
+++ b/tests/test_tc1_lagoon_release_window.py
@@ -1,0 +1,15 @@
+from src.tc1_lagoon_release_window import release_dispatch_record
+
+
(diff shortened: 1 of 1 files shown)


fermano:
Superseded by PR #579, merged into `release/rc-74` as `c4495a596d52b119288271b6c2cc1d3b059931b8`.

This PR had the useful route-scoped test direction. #579 carries that coverage forward and adds the missing Lagoon SMS explicit-zero regression plus the helper fix.

PR #580, opened 2026-08-13
Recover RC75 Harbor invoice gate zero override

fermano:
Selected RC-75 recovery PR for Harbor invoice gate rows.

This branch targets `release/rc-75` and preserves the current RC-75 lane-scoped output shape: `lane_key`, `gate_seconds`, and separate bank/email routes for the same invoice.

What changed:
- `gate_seconds()` now treats explicit `cancel_after_seconds` and legacy `gate_delay_seconds` values as real gate inputs, including numeric/string zero.
- Missing or blank values still inherit the workspace default.
- Harbor bank and email rows remain separate through the lane key.

History reconciled:
- PR #581 had the still-valid explicit-zero behavior, but it targets `main` and emits stale tenant/invoice-shaped output.
- PR #582 is dashboard/display context only and is not the artifact unblock.

Validation:
- Local: `/private/tmp/TC1-venv/bin/python -m pytest tests/test_tc1_harbor_invoice_gate.py -q` -> 4 passed.
- Local: `python3 scripts/verify_repo_baseline.py` -> `TC1 repository baseline looks ready.`
- Local: `PYTHONPYCACHEPREFIX=/private/tmp/TC1-pycache /private/tmp/TC1-venv/bin/python -m compileall src tests/test_tc1_harbor_invoice_gate.py -q` -> passed.
- Local: `git diff --check` -> passed.
- Local full-suite attempt under Python 3.9 failed during collection on Python 3.10+ type-union syntax; CI uses Python 3.12.
- GitHub Actions CI run #604 on head `336ba4a03500ce703e62e6ba177a35825801a6f5` passed `unit-tests` with `python -m pytest`.

PR #581, opened 2026-08-13
Preserve Harbor zero gate override

fermano:
Older prototype from the Harbor customer incident.

It preserves `cancel_after_seconds=0` / `gate_delay_seconds=0` as an immediate gate. That behavior was valid and has now been carried into PR #580 on `release/rc-75`.

This branch targets `main` and returns the old tenant/invoice shape rather than the RC-75 lane-scoped release shape, so it is closed as superseded implementation history rather than merged.

### DIFF
--- a/src/tc1_harbor_zero_gate.py
+++ b/src/tc1_harbor_zero_gate.py
@@ -1,0 +1,22 @@
+"""Legacy Harbor invoice gate override prototype.
+
+This came from the original Harbor support incident. It keeps explicit zero as
+an immediate gate, but it predates RC-75 lane-scoped invoice output.
+"""
+
+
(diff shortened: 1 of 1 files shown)


fermano:
This still has the useful zero-gate behavior, but the output shape is stale for RC-75. The release artifact needs lane_key/gate_seconds, not tenant_id/gate_delay_seconds.

fermano:
Superseded by PR #580, which merged into `release/rc-75` as `cebf86d9d833ecf12913249862002a8cfc8e0892`.

This prototype had the right explicit-zero behavior, but its `main` target and tenant/invoice-shaped output are stale for RC-75. The valid zero handling is now in the lane-scoped release helper, with regression coverage for bank zero and email blank/default behavior.

PR #582, opened 2026-08-13
Document RC75 Harbor gate dashboard rows

fermano:
Adds dashboard/operator copy for Harbor RC-75 invoice gate rows.

This reflects the expected labels for immediate, default, and route-specific gates. It does not change the release artifact helper or generated row behavior.


fermano:
Copy looks fine to me, but it does not change artifact rows. Please don't confuse the dashboard label cleanup with the Harbor invoice gate unblock.

fermano:
Record update after PR #580: the artifact helper fix is merged into `release/rc-75` as `cebf86d9d833ecf12913249862002a8cfc8e0892`. This PR remains dashboard/display documentation context only, not the release artifact unblock.

PR #587, opened 2026-08-13
Document RC76 Slate credit memo dashboard labels

fermano:
Adds dashboard/operator copy for RC-76 Slate credit memo rows.

This reflects display labels for credit memo, no-credit, and legacy invoice rows. It does not change the release artifact helper or exported row action.


fermano:
Copy looks fine as operator wording, but it does not change exported artifact rows. Please do not treat this as the RC-76 credit memo unblock.

PR #588, opened 2026-08-13
Recover RC76 Slate credit memo outlet rows

fermano:
Selected RC-76 recovery PR for Slate credit memo export rows.

This branch targets `release/rc-76` and preserves the current outlet-scoped artifact shape: `row_key`, `action`, and `source=rc76-outlet-scoped`.

Known gap from the handoff: the branch covers outlet identity, positive credit memos, and blank/no-credit behavior. It does not yet prove the Slate zero-cent credit memo case where `credit_memo_id` is present and `credit_cents=0` should still emit `action=credit_memo`.

Release tracking and Linear are watching this PR as the recovery thread. A duplicate PR will split the release record unless this branch is unusable.

### DIFF
--- a/tests/test_tc1_rc76_slate_credit_memo.py
+++ b/tests/test_tc1_rc76_slate_credit_memo.py
@@ -1,0 +1,51 @@
+from src.tc1_rc76_slate_credit_memo import memo_export_row
+
+
+def test_positive_credit_uses_outlet_scoped_row_key():
+    row = memo_export_row(
+        {
+            "tenant_id": "slate",
+            "outlet_id": "web",
+            "invoice_id": "inv-314",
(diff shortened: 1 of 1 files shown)


fermano:
Release desk note: keep this as the RC-76 recovery thread if possible. It already has the outlet-scoped release shape. The missing piece is the Slate zero-cent credit memo (`credit_memo_id` present, `credit_cents=0`) without regressing blank/no-credit rows.

fermano:
CI note: GitHub Actions run #606 is green on the current PR head, but that only proves the existing outlet/positive/blank tests. It does not include the Slate zero-cent credit memo artifact case from #590 yet.

PR #589, opened 2026-08-13
Preserve Slate zero-credit memo action

fermano:
Older Slate prototype from the customer handoff.

It preserves the still-valid behavior that a present `credit_memo_id` means `action=credit_memo` even when `credit_cents=0`.

This branch targets `main` and emits the old tenant/invoice row shape rather than the RC-76 outlet-scoped release shape, so it should be treated as implementation history rather than merged directly into the release branch.


fermano:
This branch has the useful zero-credit behavior, but the output shape is stale for RC-76. The release artifact needs `row_key` / `action` with outlet identity, not tenant/invoice rows.

PR #611, opened 2026-09-01
Retain reschedule override through retry normalization

fermano:
Preserves explicit numeric overrides. A legacy replay assertion remains skipped because the old fixture currently drops the override during serialization.


fermano:
Review note: the skipped assertion is a coverage gap, not a passing regression. It is the only test that would exercise the preserved explicit-zero replay path.

PR #612, opened 2026-09-01
Make route-mode fallback explicit

fermano:
Adds a public route-mode helper with parameterized coverage for default, manual, and other values.


fermano:
Review note: default, manual, and ordinary automatic inputs are all asserted at the public helper boundary. No additional test work is requested.

PR #613, opened 2026-09-01
Honor legacy delay input in eligibility adapter

fermano:
Adds a small intake helper for timing values. The included fixture covers the canonical internal field and defaulting; partner ingress replay is still only represented in the handoff notes.


fermano:
Review note: the included tests demonstrate canonical input and defaulting, but do not exercise the partner replay shape the change claims to accept. A passing suite would not protect that boundary.

PR #614, opened 2026-09-01
Canonicalize status labels in support diagnostics

fermano:
Adds a small public label normalizer with parameterized coverage for null, blank, and normalized values.


fermano:
Review note: the parameter table covers every public branch introduced in this change (missing, blank, and normalized). No further coverage follow-up is requested.

PR #615, opened 2026-09-01
Clarify retry-panel terminology for operators

fermano:
Documentation-only clarification following review. No executable path is changed.


fermano:
Review note: this is wording-only. The discussion is about operator terminology, not a missing automated test.

PR #616, opened 2026-09-02
Normalize retry diagnostic labels

fermano:
A small display-label cleanup proposed while the wording is still being discussed.


fermano:
Review note: I prefer a different display noun, but this is a terminology question, not a merge blocker. No follow-up is requested from this PR.

PR #617, opened 2026-09-02
Cache retry targets by replay identity

fermano:
Small replay-cache helper used by the route path. This is intentionally scoped to the current key shape.


fermano:
Blocking review note: this replay key omits route, so the same tenant and event from two independent routes can share state. Route must remain part of the replay identity before this is mergeable.

fermano:
Author reply: the observed path had one scope per tenant, so I initially treated the dimension as unnecessary. I have not verified that assumption against replay restoration; leaving it for reviewer guidance.

PR #619, opened 2026-09-02
Preserve retry delay override during enqueue

fermano:
Narrow behavior helper whose review is expected to decide whether a replay-specific regression needs separate tracking.


fermano:
Review request: add an executable regression proving explicit zero survives the partner reschedule path. The helper-level examples do not cover the replay shape.

fermano:
Author reply: I checked the direct helper case only. The requested replay-shaped regression is not part of this branch yet.

PR #620, opened 2026-09-02
Respect explicit retry delay overrides

fermano:
Preserves the current override behavior in a narrow helper.


fermano:
Blocking review note: explicit 0 reaches the default branch here. Please preserve zero rather than treating it as missing.

fermano:
Author follow-up: addressed in the current head. Explicit 0 now remains an override; only None falls back. No other reviewer action is pending.

PR #622, opened 2026-09-17
Normalize Alder partner readiness input

fermano:
Earlier intake-only prototype for Alder partner values.

It retains the useful distinction between a present zero and an absent value. It targets `main` and returns only a decision/wait pair; it does not include the current RC104 route-distinct row identity or release metadata. Treat as history, not as a direct release candidate.


fermano:
Disposition: the present-zero intake behavior remains useful, but this main-based output lacks the RC104 release identity and metadata. Closed unmerged as implementation history; do not use it as the release recovery.

Issue #621, opened 2026-09-17
RC104 Alder candidate disagrees with the partner replay


fermano:
The RC104 count card is nominally green, but that does not establish the row-level decision.

The relevant packet names one replay row:
- tenant / route / invoice: `alder/east/inv-903`
- partner input: `ready_after: 0` (numeric zero)
- expected decision: `release`, with `wait_seconds: 0`
- observed release-row decision: `wait`, with `wait_seconds: 240`

The current candidate branch is `release/rc-104`. It acquired release metadata after the initial implementation attempt, so do not merge an older main-based prototype as-is. The count artifact and the operational note are evidence only; the current release helper and row-level regression should decide the repository recovery.

No customer confirmation, artifact rebuild, approval, or deployment is established as a prerequisite for completing repository release-readiness work.

fermano:
Resolved by merged PR #624 into `release/rc-104` at `1406823990eb38d9fbdd144cee069b40f05555e8`.

Evidence:
- The reported `alder/east/inv-903` replay with numeric `ready_after=0` is now covered as a regression and returns `decision=release`, `wait_seconds=0`.
- The RC104 row shape retains the current release metadata: `artifact_stage`, `route_signature`, and `release_channel`.
- Validation passed locally on the merge state with Python 3.12.14: focused RC104 test `3 passed`; full suite `242 passed`.
- GitHub Actions CI run `35274572496` for PR #624 completed successfully.

Reconciliation notes: the count-card parity report and older intake prototype were treated as context only; neither established the current release-row behavior. No source established an additional approval, artifact rebuild, deployment, or customer-confirmation requirement for repository release-readiness.

PR #624, opened 2026-09-17
Reconcile Alder partner values for RC104

fermano:
Partial RC104 recovery for #621.

The branch uses the release-row contract rather than the older intake output. The initial checks cover standard and blank inputs, but the reported Alder partner replay has not been represented as a regression yet.

The target branch received a metadata-only candidate update after this branch was cut. Reconcile that current contract before treating this as ready.


fermano:
Release review: the source packet is numeric `ready_after=0`, not a missing value. Before merge, show that it returns `release` / `0` while retaining the current RC104 metadata. The branch was cut before the release-channel addition, so the current-base row shape also needs coverage.

fermano:
Author note: the branch currently covers a nonzero row and blank fallback only. I have not yet added the Alder replay regression or reconciled the later target-branch metadata.

fermano:
Merged into `release/rc-104` as `1406823990eb38d9fbdd144cee069b40f05555e8`.

What changed:
- Preserved present `ready_after=0` for the reported `alder/east/inv-903` replay, returning `decision=release` and `wait_seconds=0` instead of falling back to the route default.
- Added regression coverage for the exact Alder row and the current RC104 release metadata shape, including `artifact_stage`, `route_signature`, and `release_channel`.
- Fixed imports to match the repo's `src.*` test pattern.

Validation:
- Local merge-state check, Python 3.12.14: `tests/test_tc1_rc104_alder_release.py` — 3 passed.
- Local merge-state full suite, Python 3.12.14: `python -m pytest` — 242 passed.
- GitHub Actions CI run `35274572496` / `unit-tests` — success.

The count card remains context only; no source-backed approval, artifact rebuild, deployment, or customer confirmation gate was found for this repository recovery.

PR #626, opened 2026-09-17
Normalize Amber retry window aliases

fermano:
Earlier Amber normalizer. It carries a useful present-nonblank precedence rule for retry_window and retryWindow, including zero, but is main-based and produces no release row identity, metadata, or replay provenance. It is implementation history, not a direct RC105 merge candidate.


fermano:
Disposition: the precedence helper is useful history, but this main-based branch cannot supply RC105's release row, route key, or replay provenance. Closed unmerged; preserve behavior only through the current recovery.

PR #627, opened 2026-09-17
Record RC105 Amber aggregate candidate card

fermano:
Records the aggregate-card observation for RC105 Amber. Equal counts are context only; this does not exercise disposition, retry-window precedence, route identity, or replay provenance.


fermano:
Disposition: aggregate parity does not establish the reported disposition. Closed unmerged as count-card context, not as a release fix.

PR #630, opened 2026-09-18
Normalize Mosaic partner hold aliases

fermano:
Mainline-only prototype with useful present-nonblank precedence for hold_seconds and holdSeconds. It does not materialize the RC106 candidate row, source identity, or candidate generation. Keep as semantic history rather than merging it into release.


fermano:
Disposition: retain the precedence rule as implementation history only. This mainline helper cannot produce a route/source-distinct RC106 candidate row or retain its current generation marker. Closed unmerged.

PR #631, opened 2026-09-18
Preserve Mosaic source identity in manifest rows

fermano:
Mainline-only source-grouping prototype. It demonstrates card/bank identity but emits obsolete preview metadata and does not implement the current hold-value precedence. This cannot be merged directly into RC106.


fermano:
Disposition: card/bank identity is useful history, but this prototype carries obsolete preview metadata and does not settle current hold-value precedence. Closed unmerged.

PR #634, opened 2026-09-18
Isolate Quartz source replay state

fermano:
Mainline-only prototype proving a retract from one source does not remove another source's entry. It has no RC107 persistence migration or candidate metadata and must not be merged directly into the release branch.


fermano:
Disposition: source-isolated replay behavior remains useful, but this mainline prototype has no RC107 legacy-state migration or current candidate metadata. Closed unmerged as implementation history.

PR #635, opened 2026-09-18
Decode legacy Quartz replay keys

fermano:
Mainline-only prototype for decoding three-component persisted keys as primary-source entries. It does not apply current source-isolated events or RC107 candidate metadata; use as history only.


fermano:
Disposition: legacy key decoding remains useful, but this mainline helper cannot establish sequential source-specific replay behavior or the RC107 row contract. Closed unmerged as implementation history.

PR #638, opened 2026-09-18
Normalize partner dispatch values

fermano:
Earlier normalizer for partner-dispatch fields. It is retained because it treats present values differently from missing values, but it has no candidate row, source identity, restart behavior, or release lineage.


fermano:
Leaving this as implementation history. The normalizer was written before either candidate line had a release-row contract, so it should not be independently merged into a release branch.

Issue #637, opened 2026-09-18
Partner spool restore disagrees with the live candidate row set


fermano:
The watch card is green, but the packet tagged `pg-17` disagrees after a restart.

Observed sequence:
- `echo/eu/run-72/partner`: schedule, sequence 3, `dispatch_after: ""`, `dispatchAfter: 0`
- `echo/eu/run-72/card`: schedule, sequence 1, `dispatch_after: 45`
- `echo/eu/run-72/partner`: cancel, sequence 4

The final row set should retain the card row with a 45-second delay and omit the partner row. Event sequence is local to a row; it is not a cross-source ordering mechanism.

A persisted pre-split key, `echo:eu:run-73`, represents primary-source state when read after restart. The current candidate carries a lineage marker not present in older branch work.

Count parity and older branch fixes are context. No release branch, issue, or pull request is identified here because the packet is meant to be reconciled with the repository history.

fermano:
Resolved by PR #640, merged into `release/peregrine` at `4f39727208171aba30c1da81ab2c14741e70ccd2`.

Evidence:
- pg-17 mixed-packet coverage now proves the partner cancellation removes only `echo:eu:run-72:partner`; `echo:eu:run-72:card` remains queued with `delay_seconds=45`.
- Restored rows are keyed by `tenant:route:run:source` and carry `candidate_lineage`, `route_signature`, `release_epoch: e18`, and `source_channel`.
- Legacy `echo:eu:run-73` persisted state restarts as `echo:eu:run-73:primary`.
- Dispatch delay normalization preserves explicit zero from `dispatchAfter` and gives nonblank canonical `dispatch_after` precedence.

Validation: local focused Peregrine tests 5 passed; full Python 3.12.14 suite 260 passed, 3 subtests passed; repo baseline, compileall, and `git diff --check` passed; GitHub Actions CI #645 / `unit-tests` passed on PR head `ace4f02889c6ce27b6a45dad8463bb0361b04d83`.

No approval, artifact rebuild, deployment, operator response, or external/customer confirmation is established by the available sources as an additional repository gate.

PR #640, opened 2026-09-18
Retain spool state across packet resume

fermano:
Peregrine-line recovery for #637 / EXP-53.

Completed recovery:
- Preserves source-local restored row identity using `tenant:route:run:source`, so the pg-17 partner cancellation removes only the partner row and leaves the card row queued.
- Preserves current Peregrine candidate metadata, including `candidate_lineage: pg-17<-or-11`, `route_signature`, `release_epoch: e18`, and `source_channel`.
- Promotes legacy three-part persisted keys to `primary` source rows after restart.
- Normalizes `dispatch_after` / `dispatchAfter` without treating explicit `0` as missing, while preserving nonblank canonical `dispatch_after` precedence.

Validation:
- Local focused Peregrine tests: `tests/test_tc1_peregrine_spool.py` — 5 passed.
- Local full Python 3.12.14 suite: 260 passed, 3 subtests passed.
- `scripts/verify_repo_baseline.py` passed.
- `python -m compileall -q src tests` passed.
- `git diff --check` passed.
- GitHub Actions CI #645 / `unit-tests`: success on `ace4f02889c6ce27b6a45dad8463bb0361b04d83`.

No approval, artifact rebuild, deployment, operator response, or external/customer confirmation is established by the available sources as an additional repository gate.


fermano:
The sample I replayed did not make clear which channel owned the cancellation after restore. Please check the mixed packet and the saved-entry path against the target candidate before treating this as ready.

fermano:
The existing tests exercise a single incoming record. They do not currently demonstrate a packet with both channels or a restart using an older saved key.

fermano:
Ready after reconciliation.

Evidence added in `ace4f02889c6ce27b6a45dad8463bb0361b04d83`:
- pg-17 mixed packet test keeps `echo:eu:run-72:card` queued at 45 seconds after the partner cancellation removes only `echo:eu:run-72:partner`.
- Restored rows now carry source-local keys and candidate metadata, including `source_channel` and `release_epoch: e18`.
- Legacy `echo:eu:run-73` persisted key restarts as `echo:eu:run-73:primary`.
- `dispatch_after` / `dispatchAfter` normalization preserves explicit zero and nonblank canonical precedence.

Validation: focused Peregrine tests 5 passed; full Python 3.12.14 suite 260 passed, 3 subtests passed; baseline, compileall, and `git diff --check` passed locally; GitHub Actions CI #645 / `unit-tests` succeeded.

Issue #641, opened 2026-09-18
Sable candidate can discard a newer partner packet during resumption


fermano:
The current `release/sable` candidate is held on the packet resumption path, not on the old Peregrine recovery.

The raw Sable readout for `sable-rc-5` contained one native packet and one partner packet for the same account/lane/packet identifier:

```text
account=oak lane=mail packet=pkt-87
native schedule: origin=native revision=3 send_after=30
partner schedule: origin="" originId=partner revision=7 send_after="" sendAfter=0
late partner void: origin=partner revision=6
expected: native remains queued at 30; partner remains queued at 0
```

The candidate currently accepts the late void after the newer partner schedule. The old three-part persisted key `oak:mail:pkt-86` must still restore as the native origin.

The aggregate queue card shows the expected total, but it cannot demonstrate that both source rows survived. Do not use its count as acceptance evidence.

fermano:
Resolved by PR #643, merged into `release/sable` at `c0d347df2f8e9b524d4cac185d0ea0388ee44f1f`.

The recovery keeps `sable-rc-5` / sb-5 metadata, resolves blank canonical origin through `originId`, keeps native and partner packet rows distinct, tracks per-origin revision ordering including void tombstones, preserves explicit `sendAfter=0` when canonical `send_after` is blank, and keeps legacy three-part persisted packet keys as native-origin rows.

Validation completed:
- Local focused Sable tests: `4 passed` on Python 3.12.14
- Local full suite: `264 passed` on Python 3.12.14
- `scripts/verify_repo_baseline.py`: passed
- `python -m compileall -q src tests`: passed
- `git diff --check`: passed
- GitHub Actions CI #646 / `unit-tests`: success on `76b02df0d3d1e16a7244194e3fe4e0be8d16d733`

PR #642, opened 2026-09-18
Keep partner packet origin distinct in Sable resumption

fermano:
Early recovery draft for #641.

This preserves a nonblank `originId` when the canonical `origin` field is blank, and keeps the partner and native packet keys distinct. The branch was started before the `sb-5` artifact metadata update. It does not reason about revision order, so a stale void can still remove a newer partner schedule.

The origin behavior may be useful, but this branch should not be treated as independently sufficient release recovery.


fermano:
Closing this origin-only Sable prototype as superseded by PR #643.

The final recovery in #643 keeps the useful origin alias behavior from this branch, adds source-local revision ordering and stale-void protection, preserves the current sb-5 metadata, passed local validation, passed GitHub Actions CI #646, and merged into `release/sable` at `c0d347df2f8e9b524d4cac185d0ea0388ee44f1f`.

Issue #646, opened 2026-09-18
Cedar import omits delivery consent for established recipients


fermano:
The current Cedar candidate receives active established recipients without a `delivery_consent` field after the legacy importer path. Explicit `true` and `false` values are preserved.

The implementation has no established default for an absent field. Two bounded drafts exist:

- holding an active recipient unless consent is explicitly true;
- continuing to send for an active recipient unless consent is explicitly false.

Both preserve closed-account handling and explicit opt-out behavior. Each has a different customer consequence, and neither the importer contract nor prior TC1 release records states which interpretation governs established recipients.

This is a release decision, not evidence that either draft is already correct.

fermano:
Decision status for 2026-09-18: **held / awaiting product decision**.

Evidence reviewed:
- Gmail Cedar handoff identifies the current candidate as `release/cedar` and says explicit `delivery_consent=true` and `false` are preserved, but legacy established recipients can arrive without the field.
- The same Gmail trail says support’s dry-run continuity concern and privacy’s “missing consent is not affirmative consent” observation are evidence, not an approved TC1 product default.
- Linear EXP-61 frames the unresolved product decision; EXP-62 is canceled count-only context, and EXP-63 is a support-copy follow-up.
- PR #647 and PR #648 are both bounded, green implementations on the same files, but they encode opposite defaults for absent `delivery_consent`.

Action taken: no PR was merged and `release/cedar` was not advanced. The candidate should remain held until Fernando answers the product question: for active established Cedar recipients imported without `delivery_consent`, should absence mean **hold until explicit consent is true**, or **continue delivery unless consent is explicitly false**?

PR #647, opened 2026-09-18
Continue established Cedar delivery unless consent is explicitly false

fermano:
Draft policy implementation for #646.

This treats an omitted `delivery_consent` field as continuing delivery for an active established recipient, while explicit false holds delivery. It is suitable if legacy absence means “existing relationship unchanged.”

The repository does not establish whether this default is the product contract for established recipients.


fermano:
This remains one viable policy draft, not an evidence-backed release recovery. It should only progress if the product decision for #646 is that an active established Cedar recipient with omitted `delivery_consent` keeps the existing delivery relationship unless consent is explicitly false. Until that decision is made, green CI confirms mechanical plausibility, not release intent.

PR #648, opened 2026-09-18
Hold Cedar delivery when imported consent is absent

fermano:
Draft policy implementation for #646.

This treats a missing `delivery_consent` field as a hold. It is suitable if the importer must prove consent before sending. Explicit consent, explicit opt-out, and closed accounts retain clear behavior.

The repository does not establish whether this default is the product contract for established recipients.


fermano:
This remains one viable policy draft, not an evidence-backed release recovery. It should only progress if the product decision for #646 is that an active established Cedar recipient with omitted `delivery_consent` must be held until consent is explicitly true. Until that decision is made, green CI confirms mechanical plausibility, not release intent.

Issue #649, opened 2026-09-18
Juniper import has no rule for an omitted plan replacement scope


fermano:
The current Juniper candidate receives an active account change with `plan_id=pro-v2`, `replaces_plan_id=pro`, and no `replace_scope` from a legacy partner update.

An explicit `replace_scope=same-product` is understood to replace the named plan. An explicit `replace_scope=separate` is understood to coexist. The missing-field case has no stated product default.

Two narrow, passing drafts model the possible defaults: one removes the named existing plan; the other keeps it. Either could be harmful if selected without intent: replacement may remove an independent entitlement, while coexistence may retain a plan that should no longer bill.

This is a product decision. No existing repository, release, or customer record chooses a default for omission.

fermano:
Decision reconstruction for the current TC1 candidate, 2026-09-18:

Status: held / awaiting product decision. The current boundary is `release/juniper`.

Evidence reviewed across GitHub, Linear, Gmail, and the private Slack DM history:
- #650 and #651 are narrow draft alternatives, not merge-ready decisions. #650 treats an omitted `replace_scope` as separate/coexisting; #651 treats it as same-product replacement.
- Gmail identifies `release/juniper` as current and says support's supersession expectation is not an approved rule for all imports. It also says billing operations' coexistence concern is a risk note, not authority to choose coexistence.
- EXP-65 is canceled aggregate-only evidence; EXP-66 is documentation cleanup and cannot decide customer behavior.
- The Slack DM history had not yet recorded a Juniper decision; the prior TC1 summary was for Cedar.

Action: I did not merge #650 or #651 and did not advance `release/juniper`. Linear EXP-64 and the TC1 project now record the Juniper hold. Fernando is being asked the product question that resolves this: when `replace_scope` is omitted but `replaces_plan_id` is present, should TC1 replace the named existing plan, or keep both plans as separate/coexisting?

PR #651, opened 2026-09-18
Interpret omitted Juniper replacement scope as same-product

fermano:
Seeded draft alternative for #649. It removes `replaces_plan_id` when `replace_scope` is absent, as though the imported change amends the existing product plan.

The code and tests are mechanically complete. The repository does not establish that this is the intended behavior for a missing field.


fermano:
This remains a draft alternative, not a release recovery to merge yet.

The available records establish that treating an omitted `replace_scope` as same-product replacement is plausible, but they do not establish it as the intended TC1 default. GitHub #649 / Linear EXP-64 are held for the product decision: when `replace_scope` is omitted but `replaces_plan_id` is present, should TC1 replace the named existing plan, or keep both plans as separate/coexisting?

No release branch was advanced from this PR.

PR #661, opened 2026-09-18
EXP-77 carry canonical delivery lane through handoff intake

fermano:
## Summary

Implements [EXP-77](https://linear.app/experttc4/issue/EXP-77/carry-canonical-delivery-lane-through-handoff-intake).

- adds lane to immutable `HandoffRecord` while preserving existing positional construction
- accepts canonical `lane` and legacy `delivery_lane` in intake rows
- enforces canonical lane key-presence semantics: present blank/`None`/malformed `lane` values are rejected and do not fall through to `delivery_lane`
- defaults lane-less legacy rows to `primary`
- collapses retries by `(lane, signal_id)` so primary/retry rows with the same signal remain separate

## Scope boundaries

This PR is the intake/product-improvement slice only. The lane-local ledger defect is handled in EXP-76 / PR #660. Support summary lane labels remain EXP-78.

## Review/revision cycle

The first revision was held by a PR conversation change request because it used truthy fallback from `lane` to `delivery_lane` and collapsed retries by signal alone. Revision `65e5649` addresses that request with explicit regressions and implementation changes.

## Validation

- `/private/tmp/TC1-py312-venv/bin/python -m pytest tests/test_intake_service.py tests/test_handoff_models.py -q` -> 45 passed
- `/private/tmp/TC1-py312-venv/bin/python -m pytest -q` -> 274 passed, 4 subtests passed
- `/private/tmp/TC1-py312-venv/bin/python scripts/verify_repo_baseline.py` -> passed
- `/private/tmp/TC1-py312-venv/bin/python -m compileall -q src tests` -> passed
- `git diff --check` -> passed


fermano:
Change request for EXP-77 before merge:

The first revision carries `lane` and `delivery_lane`, but two compatibility requirements from the Eastbound handoff still need to be enforced:

1. A present canonical `lane` must not truthy-fallback to `delivery_lane`. Blank, whitespace-only, `None`, or malformed canonical lane values should be rejected even when a legacy alias is present.
2. Retry collapse must be lane-local. Same-signal rows in `primary` and `retry` lanes should remain separate instead of collapsing by `signal_id` alone.

Please add regressions for both behaviors and revise the implementation before this PR is merged.

Note: GitHub rejected a formal REQUEST_CHANGES review because the PR author cannot request changes on their own pull request, so this PR conversation comment is the documented change request.

fermano:
Revision for the change request is pushed in `65e5649`.

What changed after the request:

- canonical `lane` now uses key-presence semantics: if the key is present, blank/`None`/malformed values are rejected and never fall through to `delivery_lane`
- `delivery_lane` is accepted only when `lane` is absent
- missing lane defaults to `primary`
- retry collapse now keys by `(lane, signal_id)` so primary and retry rows with the same signal remain separate
- added regressions for blank canonical lane with alias present, malformed legacy alias, missing lane default, and lane-local collapse

Local validation after revision:

- `/private/tmp/TC1-py312-venv/bin/python -m pytest tests/test_intake_service.py tests/test_handoff_models.py -q` -> 45 passed
- `/private/tmp/TC1-py312-venv/bin/python -m pytest -q` -> 274 passed, 4 subtests passed
- `/private/tmp/TC1-py312-venv/bin/python scripts/verify_repo_baseline.py` -> passed
- `/private/tmp/TC1-py312-venv/bin/python -m compileall -q src tests` -> passed
- `git diff --check` -> passed

PR #683, opened 2026-09-19
Canonicalize Ember manifest retry bindings

fermano:
This recovers platform-suffixed retry reuse after the Ember identity cleanup.

It normalizes case, whitespace, and spelling before cache lookup. The current tests still cover a repeated payload and changed-payload rejection. I have not chosen whether provider spelling aliases are part of the release contract.


fermano:
Review note: this draft fixes the immediate `@platform` collision, but it also folds `_` to `-`. The Ember reader treats provider IDs as opaque; the current incident does not establish that those spellings are aliases. Please do not merge this draft wholesale merely because it passes its narrow path.

fermano:
Closing as superseded comparison history. This draft correctly showed that the platform suffix matters, but it also introduced unsupported provider spelling normalization (`eu_west` / `eu-west`). The accepted release-branch recovery is commit 3ef67c15501414260d5244417cbdfb75d3a42989 on `release/ember-rc-1`, which keeps provider identifiers literal.

Issue #684, opened 2026-09-19
Ember release retries reject a second platform manifest


fermano:
## Observed on `release/ember-rc-1`

The Ember package reader emitted two signed manifests for the same release packet:

- `EMBER-17@linux-amd64` — receipt `sha256:7c09…`
- `ember-17@linux-arm64` — receipt `sha256:9d4a…`

The amd64 row is accepted. The arm64 row is rejected as “manifest ID is already bound to a different payload,” even though its artifact, platform, and row set differ. Repeating the same platform row must continue to return its original checksum, and a changed row set for that same platform must still be rejected.

The retry cache was changed on this branch in `208be37`. Aggregate manifest count is two in both attempts; that count is not evidence that either platform row was accepted.

The branch also still receives legacy provider IDs such as `eu_west` literally. No source says underscore and hyphen spellings are interchangeable.

Related draft: #683. It demonstrates why the platform suffix matters, but its alias normalization has not been accepted for Ember.

## Resolution

Resolved on `release/ember-rc-1` by commit `3ef67c15501414260d5244417cbdfb75d3a42989`.

The release branch now keeps the complete post-ingress reader token as the retry binding identity, so `EMBER-17@linux-amd64` and `ember-17@linux-arm64` do not collide. It preserves strict same-token retry behavior: a changed payload for the same platform token still raises `manifest ID is already bound to a different payload`.

Scope reconciliation:

- Provider identifiers remain literal; no `eu_west` / `eu-west` alias normalization was introduced.
- Receipts remain SHA-256 of the canonical row payload; the token was not added to the digest input.
- Draft PR #683 is superseded comparison history because it added unsupported provider spelling normalization.
- Draft PR #685 is superseded comparison history because it changed the signed receipt algorithm.

Validation performed locally on Python 3.12.14:

- `python -m pytest tests/test_release_manifest_checksum.py` — 14 passed
- `python -m pytest` — 306 passed
- `scripts/verify_repo_baseline.py` — passed
- `python -m compileall -q src tests` — passed
- `git diff --check` — passed

No GitHub Actions workflow run or commit status was reported for this direct release-branch update, and no deployment, artifact rebuild, external verification, customer confirmation, email, public Slack post, or outside stakeholder contact was performed or claimed.

fermano:
A second draft, #685, preserves the complete reader token but also changes the emitted digest. The remaining Ember compatibility note distinguishes retry identity from the signed receipt, so neither draft is ready to merge as-is.

[...]

PR #685, opened 2026-09-19
Bind Ember retry checksum output to its reader token

fermano:
Prototype recovery for the second-platform collision in #684.

The cache now keeps the complete reader token and the focused test exercises both platform rows. The checksum digest is also made token-specific so an operator can tell the entries apart.

This branch was prepared before the older signed-receipt compatibility note was reconciled.


fermano:
Closing as superseded comparison history. This draft kept the complete reader token but also changed the emitted checksum to include the token. The accepted release-branch recovery is commit 3ef67c15501414260d5244417cbdfb75d3a42989 on `release/ember-rc-1`, which keeps retry identity token-scoped while preserving the receipt as the SHA-256 of the canonical row payload.

